PR #10, opened 2016-10-20
Elaborated the instructions a bit.

golddranks:
I had hard time trying to test the client, because I mixed up the general installation instructions for VSCode extensions and this README. I made some changes that elaborate things a bit.


nrc:
Thank you!

PR #14, opened 2016-12-12
Upgraded vscode dependencies

perlun:
I did this while debugging an issue I was having; unfortunately it didn't help, but at least it shouldn't make things worse (and moving onto later versions is usually a good thing).

This should be tested by someone else than me; I get errors like below both before and after this change, so it's hard for me to verify that it's fully working. But at least it builds.

```
Uncaught Exception:  Error: Header must provide a Content-Length property.
Error: Header must provide a Content-Length property.
    at StreamMessageReader.onData (/Users/per/git/rls_vscode/node_modules/vscode-jsonrpc/lib/messageReader.js:179:27)
    at Socket.<anonymous> (/Users/per/git/rls_vscode/node_modules/vscode-jsonrpc/lib/messageReader.js:164:19)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at readableAddChunk (_stream_readable.js:176:18)
    at Socket.Readable.push (_stream_readable.js:134:10)
    at Pipe.onread (net.js:543:20)
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,19 +1,19 @@
 {
-    "name": "testls-client",
-    "displayName": "testls_client",
+    "name": "rls_vscode",
+    "displayName": "rls_vscode",
     "description": "",
     "version": "0.0.1",
     "publisher": "JonathanTurner",
     "engines": {
-        "vscode": "^1.4.0"
+        "vscode": "^1.7.0"
     },
     "categories": [
         "Languages",
         "Linters",
         "Other"
     ],
-	"activationEvents": [
-		"onLanguage:rust"
+    "activationEvents": [
+        "onLanguage:rust"
     ],
     "main": "./out/src/extension",
     "contributes": {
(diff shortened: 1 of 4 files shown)


perlun:
Hmm, weird. I tried to debug this a bit more; updated the rustc nightly and so forth, but no help.

This is what it looks like:

image

When looking at the value of the `headers` object, it does indeed look a bit fishy (the `Content-Length` stuff at the end is what I'm referring to, more specifically).

image

Will gladly provide more details to anyone interested in helping out debugging this. I'm in deep water already. 😄

nrc:
This usually happens when you have a rogue `println!` writing some debugging stuff to stdout. Anything written to stdout gets sent to VS Code.

perlun:
Thanks for the merge @nrc!

> This usually happens when you have a rogue println! writing some debugging stuff to stdout. Anything written to stdout gets sent to VS Code.

Hmm, that's weird. I haven't really touched the `rls` part at all. Just compiled it. Running it from the commandline seems to work, but it's hard to test without forging a full jsonrpc message or something.

----

Like you suggested elsewhere, doing a session over IRC next week would be great; I just happen to be a bit busy right now before Christmas and all. If you want to reproduce the problem I'm seeing, you could try re-cloning `rls` and `rls_vscode`, and build it from scratch (after ensuring that your local Visual Studio Code is 1.8). If that works for you, I would be _very_ surprised!

jaxx:
I have exactly the same problem.

perlun:
@jaxx - I am suspecting the problem is in `rls` actually, or more like this: Visual Studio Code has probably changed their interpretation of the Language Server Protocol in one of their later releases. I suspect https://github.com/jonathandturner/rls/issues/103 and the subsequent https://github.com/Microsoft/language-server-protocol/issues/127 could be the trigger here...

perlun:
Hmm, weird: I skimmed through the 127 thread and didn't find anything obvious that has been changed in this area. jaxx, if you have the time, please file a new issue (in this repo) where we can track this issue. This PR is a bit hard to find, since it refers to the already-merged upgrade to newer toolchain support that I added.

jaxx:
@perlun I can file a new issue to RLS repo, yes. I'm going to install new Linux virtual machine and trying to reproduce it while writing down the steps I do.

PR #20, opened 2017-01-16
Fix #18: Count outstanding RLS analysis results to correctly show when it finishes

### LINKED ISSUE #18: Sometimes misses updates from rls
I noticed that if I'm editing, I'll sometimes miss that the rls is still working.

In this case, the vscode plugin lists the analysis as being complete when it's not.  This is mostly just confusing, as the rls/plugin do sync back up and become available again.

jonasbb:
I am not sure if it makes sense to fully encapsulate the counter as its own class. However, when testing this and only using a variable I sometimes got negative counts, so this might be prevented.

The problem occurs if outstanding requests are squashed, but the real analysis has not finished yet. Therefore, the done should only be shown if all requests finished.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -9,6 +9,26 @@ import { LanguageClient, LanguageClientOptions, SettingMonitor, ServerOptions, T
 
 let DEV_MODE = false;
 
+class Counter {
+    count: number;
+
+    constructor() {
+        this.count = 0;
+    }
+
+    increment() {
+        this.count += 1;
+    }
+
+    decrementAndGet() {
+        this.count -= 1;
+        if (this.count < 0) {
+            this.count = 0;
+        }
+        return this.count;
+    }
+}
+
 export function activate(context: ExtensionContext) {
     let serverOptions: ServerOptions;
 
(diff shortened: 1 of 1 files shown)


jonathandturner:
Thanks for the PR.

Looking over it, I couldn't see where this compensates for build tasks that are being dropped by the RLS because they've been superseded by more recent requests.

You can see this back-pressure logic here: https://github.com/jonathandturner/rls/blob/master/src/build.rs#L219-L226

nrc:
An alternative approach would be sending ids with the notifications so we can match build requests with 'done' messages and avoid counting. You would still have the same issues with builds being dropped though.

jonasbb:
I used the mentioned single variable for most of my testing on Saturday and I did not see any problems, except for sometimes negative numbers.

@jonathandturner I am not sure I understand what you mean by "compensating for build tasks that are being dropped". The problem in #18 occurs in this execution order:

* Request Build 1 => VS Code shows "RLS analysis: started"
* Request Build 2 (e.g., a second character is typed)
* RLS starts Build 1 and squashed Build 2
* VS Code received a "diagnosticsEnd" message for Build 2 => VS Code shows "RLS analysis: done"
* RLS is still working on Build 1

By counting how many outstanding build requests there are and only showing "done" when all of them are finished this solves the discrepancy of RLS state and VS Code message.

@nrc idea of having a set of outstanding requests and matching ids would work as well.

jonathandturner:
Seems good.  I'm sure we'll end up tweaking this a bit as we continue, so we can do things like show progress bars, but this gives us something to start with.

PR #22, opened 2017-01-18
Add .editorconfig

saneyuki:
* The configurations are got from exist files in this repository.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+
+[*.{ts,json,md}]
+indent_style = space
+indent_size = 4


jonathandturner:
What's the .editorconfig for?  Not sure I've used one before (maybe I have and don't remember?)

saneyuki:
@jonathandturner 

`.editorconfig` is a configuration file for [EditorConfig](http://editorconfig.org/), which specify a file formatting style for an editor which supports it via plugin for them. For example, there is [the extension for vscode](https://github.com/editorconfig/editorconfig-vscode).

I think this is useful to express and recommend the file formatting style in this project if there is no enforcement by formatting tools (e.g. `rustfmt` or `gofmt`), linter (e.g. `tslint`or `clippy`), or default conventions by a language community.

Of course, the most proper way is an enforcement by CI or formatting tools, but this project has not had it yet. And there may not be good tools to lint for Markdown or json file (at lease, I don't know proper de-facto tools). So I think EditorConfig is useful as loose recommendations for this project at this time.

nrc:
Thanks for the explanation!

Llonger term we should have rustfmt support in the CI (well, I guess prettier or something for the JS/TypeScript stuff). I think it is better for editor-specifc config files such as this to live out of tree.

PR #29, opened 2017-01-19
Add manual VSCode extension installation info

machineloop:
Thanks for the fantastic work, just wanted to add a few notes on how to manually install the extension to try on multiple Rust projects without using the debugger. Also added a link to the VSCode extension docs. Hopefully this can save someone else a few minutes of time.

Thanks again, this is an extremely exciting first look at where the rust tooling is headed!

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,3 +19,16 @@ variable to the root of your rls checkout:
 ```
 export RLS_ROOT=/Source/rls
 ```  
+You can also add this export to your bash profile or equivalent.
+
+## Manually add as regular VSCode extension
+
+If you'd like to test on multiple projects and already have the extension working properly, you can manually install the extension so that it's loaded into VSCode by default.
+
+After following the above instructions, and successfully building the extension once, symlink or copy the `rls_vscode` directory to either:
+```
+Windows: %USERPROFILE%\.vscode\extensions
+Mac/Linux: $HOME/.vscode/extensions
+```
+For example, to setup a symlink on Mac/Linux: `ln -s /path/to/rls_vscode/ ~/.vscode/extensions/rls_vscode`
+Restart VSCode in order to load the extension. More information available via [VSCode docs](https://code.visualstudio.com/Docs/extensions/example-hello-world#_installing-your-extension-locally).


README.md
```diff
@@ -19,3 +19,13 @@ variable to the root of your rls checkout:
 ```
 export RLS_ROOT=/Source/rls
 ```  
+## Manually add as regular VSCode extension
+
+If you'd like to test on multiple projects and already have the extension working properly, want to load the extension by default.
```
nrc: I think there is a grammar mistake here, needs some words before "want"

README.md
```diff
@@ -19,3 +19,13 @@ variable to the root of your rls checkout:
...
+
+After following the above instructions, and successfully building the extension once, copy the entire `rls_vscode` directory to either:
+```
+Windows: %USERPROFILE%\.vscode\extensions
+Mac/Linux: $HOME/.vscode/extensions
+```
+You can also add `export RLS_ROOT=/Source/rls` to your bash profile or equivalent. More information available via [VSCode docs](https://code.visualstudio.com/Docs/extensions/example-hello-world#_installing-your-extension-locally).
```
nrc: This note on `RLS_ROOT` feels like it belongs to the above section rather than this one.

README.md
```diff
@@ -19,3 +19,13 @@ variable to the root of your rls checkout:
 ```
 export RLS_ROOT=/Source/rls
 ```  
+## Manually add as regular VSCode extension
+
+If you'd like to test on multiple projects and already have the extension working properly, want to load the extension by default.
+
+After following the above instructions, and successfully building the extension once, copy the entire `rls_vscode` directory to either:
```
nrc: Do I need  to restart VSCode after doing this or does it just detect the change?

README.md
machineloop: Thanks, good catch, clarified in additional commit.

README.md
machineloop: Yes, thanks, restarting VSCode is necessary, added in commit

README.md
machineloop: Moved to above section below export snippet

machineloop:
Thanks for the feedback on the notes, fixed in additional commit. Let me know if you think the wording could use some more work. Thanks again for working on this!

chriskrycho:
One thing worth noting—I believe you should be able to symlink it instead of copying it, which would have the upside of not requiring you to copy-and-paste again every time you update the extension. (I've done this with my own small VS Code extension, so I know it works).

machineloop:
Thanks @chriskrycho, good point. I've changed line to:
>symlink or copy the `rls_vscode` directory

And also added example unix symlink command:

>For example, to setup a symlink on Mac/Linux: `ln -s /path/to/rls_vscode/ ~/.vscode/extensions/rls_vscode`

Squashed to single commit.

nrc:
Thanks!

PR #39, opened 2017-03-22
Add output channel

shakram02:
the output channel will make seeing what's happening with RLS much easier, and it's also not intrusive for other users

image


nrc:
This shouldn't be necessary. If you run in debug mode you already see this in the extensions/RLS view.

shakram02:
would you please tell me how to do it ? because when I use the extension RLS just works, I can't see any trace messages or so

nrc:
Set `DEV_MODE` to `true` here: https://github.com/jonathandturner/rls_vscode/blob/master/src/extension.ts#L10

PR #40, opened 2017-03-26
Make spinner consistent despite multiple running diagnostics

Xanewok:
I noticed that sometimes when modifying code the spinner behaved erratically - it could sometimes appear as stuck or be spinning very, very fast. Upon further inspection it seems that it's related to multiple timers updating status bar message at at different times.
Making it use at most 1 timer seems to have fixed the problem. From what I read about Node.JS architecture/intervals, access to spinnerTimer should be safe, as there's only single thread executing interval callbacks.
I'm not a JS/JS developer myself, so I'm open to any suggestions regarding this change. It's a minuscule change, but I hope it still has some value :)

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -9,7 +9,8 @@ import { LanguageClient, LanguageClientOptions, SettingMonitor, ServerOptions, T
 
 let DEV_MODE = false;
 
-let spinnerTimers = [];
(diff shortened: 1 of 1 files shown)


nrc:
LGTM, but r? @jonathandturner

jonathandturner:
Thank you!

PR #49, opened 2017-04-18
Nicer error message if rls process could not be spawned

dbrgn:
- Catch ENOENT errors
- Print nicer error log than default
- Update the status bar with an error message

Without this, the user thinks that the RLS is still starting up.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -54,7 +54,17 @@ export function activate(context: ExtensionContext) {
                 } else {
                     childProcess = child_process.spawn("rls");
                 }
+
                 childProcess.stderr.on('data', data => {});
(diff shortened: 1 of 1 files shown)


jonathandturner:
Cool, nice fix!

I was wondering, should we detect if `rls` is available also?  That way we can say "did you install the 'rls' command?"

dbrgn:
Yeah, but that should probably be done *before* trying to launch the binary :)

Xanewok:
An idea - using the same listener mechanism we could additionally catch `'close'`/`'error'` and set appropriate status message (such as `process quit with code {}`), so the user knows when the analysis process crashed.

PR #50, opened 2017-04-22
Remove "hello world" contribution

roblourens:
This adds a "hello world" command to the command palette which fails because the command doesn't exist.


jonathandturner:
Thanks!

PR #58, opened 2017-06-07
Handle process spawning error regardless of DEV_MODE

Xanewok:
Made so that serverOptions are being initialized regardless of DEV_MODE, which reduces code duplication a little bit and does error handling regardless of DEV_MODE setting.
This is helpful for working on RLS, since the errors aren't swallowed (status bar and log) with DEV_MODE on.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -11,7 +11,6 @@ let DEV_MODE = false;
 
 let spinnerTimer = null;
 let spinner = ['|', '/', '-', '\\'];
-let nextBuildTask = 0;
 
 class Counter {
     count: number;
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -34,43 +33,34 @@ class Counter {
...
-            function spawnServer(...args: string[]): child_process.ChildProcess {
-                let childProcess;
-                if (rls_root) {
-                    childProcess = child_process.spawn("cargo", ["run", "--release"], { cwd: rls_root });
+        function spawnServer(...args: string[]): child_process.ChildProcess {
+            let childProcess = rls_root ? child_process.spawn("cargo", ["run", "--release"], { cwd: rls_root })
+                                        : child_process.spawn("rustup", ["run", "nightly", "rls"]);
+
+            childProcess.on('error', err => {
+                if ((<any>err).code == "ENOENT") {
```
Xanewok: Initializing `childProcess` with ternary properly infers a type for it, but because of that `tsc` expects `err` to be generic `Error` here, which is overriden in node. Couldn't find proper TS interface for that (`code` is specific to node here), so in this place I'm casting it to `any` (as it worked before) to avoid type checking.

Xanewok:
Rebased after https://github.com/jonathandturner/rls_vscode/pull/59.

nrc:
Looks good, thank you!

Issue #66, opened 2017-06-21
stderr not being written to console


nrc:
Even with `HIDE_WINDOW_OUTPUT = false`

Xanewok:
Asked also over at https://github.com/Microsoft/vscode-languageserver-node/issues/223

nrc:
Even after #68, this is not working for me.

nrc:
Fixed now

Xanewok:
Workaround is not needed as of https://github.com/Microsoft/vscode-languageserver-node/commit/8ed63637d95e1dbd7ad0b6e5d71195a7a88943b2 !

PR #68, opened 2017-06-27
Append to output window if not HIDE_WINDOW_OUTPUT

Xanewok:
This tries to work around weird issue https://github.com/rust-lang-nursery/rls-vscode/issues/66. Initially tested it and it looks like it should be working, but had some consistency issues with the change. If anyone knows how to do it better, please do (tell). 👍

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -13,8 +13,9 @@
 import * as child_process from 'child_process';
 import * as fs from 'fs';
 
-import { workspace, ExtensionContext, window, commands } from 'vscode';
+import { workspace, ExtensionContext, window, commands, OutputChannel } from 'vscode';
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -83,6 +87,11 @@ export function activate(context: ExtensionContext) {
 
             if (HIDE_WINDOW_OUTPUT) {
```
nrc: Can we move this whole block to the `lc.onReady` closure?

Xanewok: We won't have access to `childProcess.stderr` there and generally we won't be able to access `LanguageClient._childProcess`, because it's private and also [not assigned](https://github.com/Microsoft/vscode-languageserver-node/blob/master/client/src/main.ts#L178) (case where we spawn server with a closure)

nrc: OK, thanks for clarifying!

PR #73, opened 2017-07-17
Read extension configuration from `contributes.configuration`

Xanewok:
In short this removes hardcoded `HIDE_WINDOW_OUTPUT` and `LOG_TO_FILE` and reads it from settings.json and clarifies when the output channel will be hidden.

For the context, `OutputChannel` will always show up in the drop-down list if at least there will be one call to `OutputChannel.append`, which can happen regardless if we redirect RLS stderr to it, e.g. when receiving LSP error log message, so I separated `HIDE_WINDOW_OUTPUT` into `showStderrInOutputChannel` and `revealOutputChannelOn` configuration options.

I'm open for changing the configuration names' prefixes from "rls.client" to something else, like the other - "rust". Did it because I wanted to separate what's a configuration for the client, and what's a configuration for the server (which, for the server, right now pulls everything with "rust" prefix, [here](https://github.com/rust-lang-nursery/rls-vscode/compare/master...Xanewok:configuration?expand=1#diff-45327f86d4438556066de133327f4ca2R101), and sends it via protocol on change of it)

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,9 +5,9 @@
     "version": "0.1.0",
     "publisher": "rust-lang",
     "icon": "rust-icon.png",
-	"galleryBanner": {
-		"color": "#336356",
-		"theme": "dark"
+    "galleryBanner": {
+        "color": "#336356",
+        "theme": "dark"
     },
     "engines": {
         "vscode": "^1.8.0"
@@ -75,6 +75,27 @@
             "type": "object",
             "title": "Rust configuration",
             "properties": {
+                "rust-client.showStdErr": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "Specifies whether RLS stderr will be shown in the output channel along with protocol logs. Requires reloading extension after change."
+                },
+                "rust-client.logToFile": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "When set to true, RLS stderr is logged to a file at workspace root level. Requires reloading extension after change."
+                },
+                "rust-client.revealOutputChannelOn": {
+                    "type": "string",
+                    "enum": [
+                        "info",
+                        "warn",
+                        "error",
+                        "never"
+                    ],
+                    "default": "never",
+                    "description": "Specifies message severity on which the output channel will be revealed. Requires reloading extension after change."
+                },
                 "rust.sysroot": {
                     "type": ["string", "null"],
                     "default": null,
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -1,0 +1,52 @@
+// Copyright 2017 The RLS Developers. See the COPYRIGHT
+// file at the top-level directory of this distribution and at
+// http://rust-lang.org/COPYRIGHT.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+'use strict';
+
+import { workspace, OutputChannel, WorkspaceConfiguration } from 'vscode';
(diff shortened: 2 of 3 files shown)


nrc:
Hmm, I would certainly prefer `rust.client` to `rls.client` (I am trying to avoid mentioning the RLS except where necessary when discussing the implementation). Perhaps `rust-client.`?

package.json
```diff
@@ -74,6 +74,27 @@
             "type": "object",
             "title": "Rust configuration",
             "properties": {
+                "rls.client.showStderrInOutputChannel": {
```
nrc: Maybe just `showStdErr`

package.json
```diff
@@ -74,6 +74,27 @@
...
+                },
+                "rls.client.revealOutputChannelOn": {
+                    "type":"string",
+                    "enum": [
+                        "info",
+                        "warn",
+                        "error",
+                        "never"
+                    ],
+                    "default": "error",
```
nrc: Could default be `never`?

src/extension.ts
```diff
@@ -12,16 +12,16 @@
...
 import { runRlsViaRustup } from './rustup';
 import { startSpinner, stopSpinner } from './spinner';
+import { RLSConfiguration } from "./configuration";
 
 import * as child_process from 'child_process';
 import * as fs from 'fs';
 
-import { workspace, ExtensionContext, window, commands, OutputChannel } from 'vscode';
-import { LanguageClient, LanguageClientOptions, SettingMonitor, ServerOptions, TransportKind, NotificationType } from 'vscode-languageclient';
+import { workspace, ExtensionContext, window, commands, OutputChannel, WorkspaceConfiguration } from 'vscode';
```
nrc: Doesn't look like WorkspaceConfiguration is used?

src/configuration.ts
```diff
@@ -0,0 +1,55 @@
...
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+'use strict';
+
+import { workspace, OutputChannel, WorkspaceConfiguration } from 'vscode';
+import { RevealOutputChannelOn } from 'vscode-languageclient';
+
+export { RevealOutputChannelOn } from 'vscode-languageclient';
```
nrc: Unnecessary, I think

package.json
Xanewok: Set to `error` because it's the default level when not nothing is specified in the `ClientOptions` for the vscode-languageclient, but we can stick with `never` ourselves

src/extension.ts
Xanewok: Oops, you're right, thanks!

Xanewok:
On a second thought, I guess it doesn't matter much if the server will receive notification regarding changing configuration for the client, at least for the time being. Since the convention seems to be using commas, how about `rust.client.*`?

package.json
Xanewok: How about `showServerStdErr` to emphasize that it's server's stderr that will be shown?

src/configuration.ts
Xanewok: It's not strictly required, but since it consumes `RevealOutputChannelOn` from somewhere else and exports a function that contains this type, I thought it's good style to also export what's in the interface; I can remove it since it's only used locally in practice

nrc:
> On a second thought, I guess it doesn't matter much if the server will receive notification regarding changing configuration for the client, at least for the time being

I think it might matter because it means we will do a cargo build that could be avoided otherwise

Xanewok:
[Notification](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#didchangeconfiguration-notification) should only send changed data, however sending it at all indeed sounds wasteful.


@nrc Is `rust-client.*` okay?

nrc:
> Is rust-client.* okay?

I think so, yes

Xanewok:
Pushed and rebased with a change to `rust-client`

nrc:
Thanks for the changes!

Issue #79, opened 2017-07-17
Rust-specific VSCode commands should be prefixed with "Rust: "


Xanewok:
Many other language support extensions do that, e.g. CSharp, .NET, C/Cpp. Having such prefix would also help with finding related commands in the Command Palette.

nrc:
Addressed by #80

PR #80, opened 2017-07-17
Add "Rust" category to "Deglob imports" command

Xanewok:
Related to https://github.com/rust-lang-nursery/rls-vscode/issues/79.

This adds "Rust: " prefix to the displayed command name in the UI and logically groups related commands.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,8 @@
         ],
         "commands": [{
             "command": "rls.deglob",
-            "title": "Deglob imports"
+            "title": "Deglob imports",
+            "category": "Rust"
         }],
         "configuration": {
             "type": "object",


nrc:
Do we need to update the README too? (https://github.com/rust-lang-nursery/rls-vscode#commands)

Xanewok:
From what I checked the "Rust" category is only related to the presentation layer and just adds a namespace for command names displayed in the list.

Regarding updating the README, I think it'd be good to switch from technical commands, i.e. `deglob`, to something user can access, like `Rust: Deglob imports` in this case. Microsoft's Go support extension page for reference: https://github.com/Microsoft/vscode-go#commands.

@nrc should I update the README?

nrc:
Thanks for the explanation. I do think we should make the README more friendly, but it doesn't need to block this PR - I can do it later.

Issue #81, opened 2017-07-17
RLS analysis makes everything on the bottom bar jiggle


sunjay:
Example:
![peek 2017-07-17 10-26](https://user-images.githubusercontent.com/530939/28272770-7b9d6508-6ada-11e7-9ac7-dec040ce6df0.gif)

Would it be possible to use something fixed width or just remove the loading spinner altogether? The constant motion is very distracting.

jonathandturner:
I wonder if the font for the bottom can be set to fixed width

Xanewok:
Not sure if this is currently possible. It's possible to change the color of the statusbar, embed [octicons](octicon.github.com) in status bar text, but changing font is something that's still planned, if I'm not mistaken ([relevant](https://github.com/Microsoft/vscode/issues/26128) [issues](https://github.com/Microsoft/vscode/issues/519)).

We could technically switch to `['/', '\\']` spinner though 😛

sunjay:
This icon looks like a spinner: https://octicons.github.com/icon/sync/

Xanewok:
It does indeed, but my first impression looking at the icon is that it's related to either synchronization or reloading and we can't animate it, which I believe is important from UX standpoint to show user that's something being actively processed, so I don't believe it's a good idea to replace current spinner with this icon.

I think we should investigate this a bit, maybe toying with whitespaces will do the trick. If not, I guess we could settle for poor man's 2-character spinner and eventually wait for font theme support to land on vscode side.

Xanewok:
Then again maybe you're right and we could try and experiment with:
1. https://octicons.github.com/icon/zap/
2. https://octicons.github.com/icon/sync/
3. https://octicons.github.com/icon/check/

for the current `starting up` > `working` > `done` cycle

nrc:
Where does the `--NORMAL MODE--` come from? I think this would be fine if there weren't text to the right of the spinner (which we avoid in the extension).

sunjay:
@nrc That's from the vim mode extension. I don't think you have full control over that because I see the items on the bottom bar switch position all the time. Seems like whatever is last added or changed always gets moved to the right. (I just started using vscode from atom so this all seems like pretty weird behavior to me.)

nrc:
I would prefer to use a spinner rather than the octirons, for one thing VSCode uses a similar ascii spinner in some other situations. I wonder if we could use some fancy unicode characters that always render fixed width?

Xanewok:
[...]

PR #84, opened 2017-07-17
Use fixed-width spinner in the status bar

Xanewok:
Just an example what kind of spinner could work for https://github.com/rust-lang-nursery/rls-vscode/issues/81.

Used these characters: ◐◓◑◒


nrc:
Let's try it!

PR #77, opened 2017-07-17
Add .travis.yml to prepare to enable TravisCI

saneyuki:
- This pull request contains the change to upgrade the supported version for vscode to make ci more precise. It might has some risks, but I feel it would not be a large problem because vscode update itself automatically.
- ...And I'm not a people of [rust-lang-nursery](https://github.com/rust-lang-nursery). So someone need to enable TravisCI for this repository.

PR #85, opened 2017-07-18
Add TravisCI badge to README

saneyuki:
- This is follow up for https://github.com/rust-lang-nursery/rls-vscode/pull/77
- This badge would be turned on if we activate TravisCI for this repository (currently it is disabled.)


saneyuki:
@nrc r?

PR #86, opened 2017-07-18
Bump up TypeScript to 2.4 and enable some compiler options for type-safe

saneyuki:
- [By this change](https://github.com/saneyuki/rls-vscode/commit/e6db20d44ad93a7aea7133ef456129f0d831e607), typescript compiler bans if the source includes unused arguments. However, we can avoid the error with `_` prefix for them.
- This enables [`strict` option](https://www.typescriptlang.org/docs/handbook/compiler-options.html) for typescript compiler. It useful to make our codebase more solid.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
         "vscode-languageclient": "^3.3.0"
     },
     "devDependencies": {
-        "typescript": "^2.0.3",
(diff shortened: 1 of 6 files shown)


nrc:
Nice, thank you!

PR #90, opened 2017-07-20
Remove RevealOutputChannelOnUtil namespace

saneyuki:
namespace is TypeScript only feature and it's transformed to 'namespace object' pattern for JavaScript.

It was popular and useful for classic browsers. but for nodejs, we should use module per file.


nrc:
Thanks!

PR #91, opened 2017-07-20
Proposal:  Use yarnpkg to lock our dependencies

saneyuki:
This change switch from npm to [yarnpkg](https://yarnpkg.com/).

An extension is a kind of self-contained application unlike a library.
So I think it's reasonable to add the lock file into the repository
for dependency management.

Alternative Approach
--------------------

__Use npm v5 or later__

This plan don't requires to install yarnpkg for our contributor like this pull request.

However, it specifies npm version. npm is bundled with nodejs.
I'm not sure to that it's safe to install some specific version of npm
into user local machine.

And I also think it's a little hard to specify the version of npm
because nodejs also update bundled npm on upgrading nodejs.

If we can use nodejs 8 or later, this plan might be reasonable because
nodejs 8 bundles npm v5.x. But we use nodejs 7. Thus I propose
to use the latest yarnpkg _at this time_.


saneyuki:
- installing the latest npm (v4.x -> v5.x) requires [about 30sec](https://travis-ci.org/rust-lang-nursery/rls-vscode/builds/255590548).
- installing the latest yarnpkg requires [about 2sec](https://travis-ci.org/rust-lang-nursery/rls-vscode/builds/255592178)

By these results, I think there are additional reason to use yarnpkg to reduce the time to testing on CI.

saneyuki:
@nrc how do you think about this?

chriskrycho:
Note that npm@5 is bundled with Node 8 and later, but installing npm@5 does not depend on installing Node 8 in any particular way. Using npm@5 vs. using yarn seems to be roughly the same tradeoff in terms of installing non-bundled software.

(Qualification: I dramatically prefer Yarn to npm, and think this project *should* use Yarn. This is just adding context on this specific point.)

nrc:
Thoughts:
* for a random JS project I would use Yarn and generally approve
* for VSCode extensions, the consensus seems to be to use npm without Yarn. That is the case in the tutorials and examples, in all (I think) extensions from MS and most community extensions. I think in general, it is better to align with the community rather than use an unusual (albeit better) solution.
* I'm not really clear on the problem that is being solved here - are we trying to make things easier for contributors? Are there problems with npm versioning (beyond Linux distros shipping a ridiculously old version). Are there problems for users of the extension? Or is this purely to improve the CI?

saneyuki:
I'm sorry to this reply was late.

> I'm not really clear on the problem that is being solved here - are we trying to make things easier for contributors? Are there problems with npm versioning (beyond Linux distros shipping a ridiculously old version). Are there problems for users of the extension? Or is this purely to improve the CI?

This was my mistake which lacks to explain the core motivation about this proposal. Sorry.

I proposed this to aim to make dependencies more deterministic for contributors.

TypeScript compiler sometimes causes breaking changes on _minor_ upgrading, and `@types/***` packages' _patch/minor_ upgrading sometimes contains changes which causes compile errors. These causes un-deterministic problem on contributing to this project. And this un-deterministicness often be a problem on random failure on CI. To resolve this un-deterministic problem, npm@5 or later generates `package-lock.json` to locking dependencies.

--------------------------

I chose yarnpkg [by the above reasons](https://github.com/rust-lang-nursery/rls-vscode/pull/91#issue-244299585) on opening this proposal. But I don't have any strong opinion to avoid npm@5. I'll open a new pull request to use npm@v5  and will close this after opening new one.

saneyuki:
> I'll open a new pull request to use npm@v5 and will close this after opening new one.

I opened the new pull request: https://github.com/rust-lang-nursery/rls-vscode/issues/97

PR #97, opened 2017-07-26
Proposal: Lock our dependencies to make CI more deterministic

saneyuki:
This retakes https://github.com/rust-lang-nursery/rls-vscode/pull/91

## Motivation

TypeScript compiler sometimes causes breaking changes on minor upgrading, and @types/*** packages' patch/minor upgrading sometimes contains changes which causes compile errors. These causes un-deterministic problem on contributing to this project. And this un-deterministicness often be a problem on random failure on CI.

To resolve this un-deterministic problem, npm@5 or later generates package-lock.json to locking dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,5 +9,11 @@ language: node_js
 node_js:
   - 7.4 # Sort with the version bundled in supported vscode.
 
+before_install:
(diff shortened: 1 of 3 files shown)


nrc:
Thank you!

PR #104, opened 2017-07-31
Accept hidden "rls.path" and "rls.root" user settings

Xanewok:
With this, specified `"rls.path"` user setting should work the same as if passed RLS_PATH env var, same with `"rls.root"`. However env vars still have bigger priority than user settings.

These options are intentionally not specified in `package.json`, since they're dev-only and would confuse the user. This allows to dynamically specify which RLS to run, which is useful when working on the RLS as well.


nrc:
Do we need to add these to package.json too? Or is that what makes them hidden? I guess it doesn't matter if we don't get informed about changes, because we will restart anyway.

Xanewok:
> Do we need to add these to package.json too?

That's what makes them hidden, but they're still read and recognized if set
in the settings section

PR #117, opened 2017-08-09
fix typo in rust-analysis to get auto install working

wraithan:
This caused the extension to fail mysteriously when installing components. Error handling here should likely be improved.


nrc:
Thank you!

PR #118, opened 2017-08-12
Add problem matchers to highlight compilation errors in editor view/minimap.

jTitor:
It looks like the current build/test/check tasks don't supply a problem matcher, so errors are displayed in the output tab but aren't highlighted in the main editor/minimap. This matcher can't quite match notes, and I didn't want to switch output to JSON, so this change only matches errors and warnings.


nrc:
Awesome, thank you!

PR #123, opened 2017-08-17
Offer find implementations in the menus 

jonasbb:
Since my PR just landed for rls I want to remind you that I also have a small modification to the extension. It provides all the glue code to invoke find_impls from Vscode and adds a context menu entry.

If you dislike some of the choices or want to add something (e.g keybindings) feel free to either just push to the PR or cherry-pick and modify it.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -17,8 +17,9 @@ import { RLSConfiguration } from "./configuration";
 import * as child_process from 'child_process';
 import * as fs from 'fs';
 
-import { workspace, ExtensionContext, window, commands, OutputChannel } from 'vscode';
-import { LanguageClient, LanguageClientOptions, ServerOptions, NotificationType, RevealOutputChannelOn } from 'vscode-languageclient';
+import { workspace, ExtensionContext, TextEditor, TextEditorEdit, window, commands, OutputChannel } from 'vscode';
+import { LanguageClient, LanguageClientOptions, Location, NotificationType, RevealOutputChannelOn,
+    ServerOptions } from 'vscode-languageclient';
 import * as is from 'vscode-languageclient/lib/utils/is';
 
 const CONFIGURATION = RLSConfiguration.loadFromWorkspace();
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -158,14 +159,26 @@ function diagnosticCounter(lc: LanguageClient) {
...
+    let cmdDisposable = commands.registerTextEditorCommand('rls.deglob', (textEditor, _edit) => {
         lc.sendRequest('rustWorkspace/deglob', { uri: textEditor.document.uri.toString(), range: textEditor.selection })
             .then((_result) => {},
                   (reason) => {
                 window.showWarningMessage('deglob command failed: ' + reason);
             });
     });
     context.subscriptions.push(cmdDisposable);
+
+    cmdDisposable = commands.registerTextEditorCommand('rls.findImpls', (textEditor: TextEditor, _edit: TextEditorEdit) => {
```
nrc: Could we create a new disposable object rather than recycling the existing one please? I fear subtle breakage due to change or JS semantics or something.

src/extension.ts
```diff
@@ -158,14 +159,26 @@ function diagnosticCounter(lc: LanguageClient) {
...
+        console.log('Trace find implementations');
```
nrc: Could you remove this line please?

src/extension.ts
```diff
@@ -158,14 +159,26 @@ function diagnosticCounter(lc: LanguageClient) {
...
+        let params = lc.code2ProtocolConverter.asTextDocumentPositionParams(textEditor.document, textEditor.selection.active);
+        let respone = lc.sendRequest("rustDocument/implementations", params);
```
nrc: typo: response

jonasbb:
@nrc Thank you for the review. I addressed your comments.

src/extension.ts
jonasbb: Here I am just reusing the variable name, not the object itself. But I'll change it.

nrc: To clarify, I think this is OK now, but could be fragile. For example, if we introduce a closure here we might end up pushing the second object twice and skipping the first (I admit this is unlikely, but the alternative is almost as easy).

nrc:
Thank you!

PR #126, opened 2017-08-20
Make it possible to opt-out to add default build tasks to task.json on opening a project

saneyuki:
Motivation
----------------

The previous behavior always create `.vscode/tasks.json` if there is no it.

If the workspace does not contains `.vscode/tasks.json` intentionally by
some reasons (it is not a root dir of the repository, the repository
has not ignored `.vscode` yet, or others), then the previous behavior
force to users to ignore/remove `.vscode/tasks.json` on creating their changesets for the repository.
I'd like to stop it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ following commands:
 * `deglob` - replace a glob (`*`) import with an explicit import. E.g., replace
   `use foo::*;` with `use foo::{bar, baz};`. Select only the `*` when running
   the command.
+* `Configure default build tasks` - create `.vscode/tasks.json` with some build tasks.
 
 
 ### Snippets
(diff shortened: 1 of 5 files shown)


nrc:
I think that users should opt-out of this behaviour, rather than opt-in to it (because the opt-in is not very discoverable, and for most projects for most users, it is desirable behviour to create tasks.json). So instead of adding a command to do this could you add a config option (default to true) for the behaviour?

saneyuki:
@nrc

> So instead of adding a command to do this could you add a config option (default to true) for the behaviour?

Okay.  I'll add a new option and make this to opt-out behavior. 

BTW, I'd like to keep the change which adds a new command to setting up default tasks for all users. This command can help to update old ones with the newer tasks configured by rls  for the future, or recover default task configurations.

nrc:
> I'd like to keep the change which adds a new command to setting up default tasks for all users.

SGTM, thanks!

saneyuki:
@nrc updated!

nrc:
Great, thank you!

PR #127, opened 2017-08-21
Various clean up for default tasks

saneyuki:
Now, we can access to the problem matchers via `$rustc` alias. This is useful to reuse it for user defined tasks!


saneyuki:
@nrc r?

nrc:
Thanks!

PR #131, opened 2017-08-22
Proposal: Use workspace.registerTaskProvider() to provide default tasks

saneyuki:
## Summary

I found vscode extension API provides [`workspace.registerTaskProvider()`](https://code.visualstudio.com/docs/extensionAPI/vscode-api) to register tasks which are shown if we run "Tasks: Run Tasks" from vscode's command palette.

If we use it, we can do these:

- Provide more seemless default  tasks.
- Remove `rust-client.setup_build_tasks_automatically` introduced in https://github.com/rust-lang-nursery/rls-vscode/pull/126
    - We did not need introduce it If I found `workspace.registerTaskProvider()` when I had worked on it. Sorry.
- Remove "Add default tasks to tasks.json" command  introduced in https://github.com/rust-lang-nursery/rls-vscode/pull/126
    - We can setup them from:
        1. Do "Tasks: Run Tasks" command
        2. Select "tasks name" and click the configuration (gear) icon.

### DIFF
--- a/src/tasks.ts
+++ b/src/tasks.ts
@@ -19,60 +19,8 @@ import {
     TaskRevealKind,
     ShellExecution,
     ShellExecutionOptions,
-    window,
     workspace,
-    WorkspaceConfiguration,
 } from 'vscode';
-
-function getConfiguration(): { config: WorkspaceConfiguration; hasOtherTasks: boolean } {
-    const config = workspace.getConfiguration();
-    const hasOtherTasks: boolean = !!config['tasks'];
-
-    return {
-        config,
-        hasOtherTasks,
-    };
-}
-
-export async function addBuildCommandsByUser(): Promise<string | undefined> {
-    const { config, hasOtherTasks } = getConfiguration();
-    if (hasOtherTasks) {
-        return Promise.resolve(window.showInformationMessage('tasks.json has other tasks. Any tasks are not added.'));
-    }
-
-    return addBuildCommands(config);
-}
-
-async function addBuildCommands(config: WorkspaceConfiguration): Promise<string | undefined> {
-    try {
-        const tasks = createDefaultTaskConfig();
-        await Promise.resolve(config.update('tasks', tasks, false));
-    }
-    catch (e) {
-        console.error(e);
-        return Promise.resolve(window.showInformationMessage('Could not update tasks.json. Any tasks are not added.'));
-    }
-
-    return Promise.resolve(window.showInformationMessage('Added default build tasks for Rust'));
-}
-
-function createDefaultTaskConfig(): object {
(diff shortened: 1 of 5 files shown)


saneyuki:
@nrc 

How do you think about this?

nrc:
This sounds great, nice find!

src/tasks.ts
```diff
@@ -91,3 +105,146 @@ function createDefaultTaskConfig(): object {
...
+                type: 'shell',
+                command: 'cargo',
+                args: [
+                    'build'
+                ],
+            },
+            problemMatcher,
+            group: TaskGroup.Build,
+            presentationOptions,
+        },
```
nrc: Do we still get to be 'the' build task with this approach? I.e., do we get to call this task on ctrl + shift + b ?

nrc: I think we should only have one task in the `build` group to do this.

src/tasks.ts
```diff
@@ -91,3 +105,146 @@ function createDefaultTaskConfig(): object {
...
+        {
+            definition: {
+                taskName: 'cargo clean',
+                type: 'shell',
+                command: 'cargo',
+                args: [
+                    'clean'
+                ],
+            },
+            group: TaskGroup.Clean,
```
nrc: I don't think `clean` exists

src/tasks.ts
saneyuki: @nrc 
> do we get to call this task on ctrl + shift + b ?

Yep. This change shows up 'cargo run' or 'cargo build' if we call ctrl + shift + b.

>  I think we should only have one task in the build group to do this.

okay. I'll remove it from `cargo run`.

src/tasks.ts
saneyuki: @nrc 

>  I don't think clean exists

The current vscode does not has a equivalent way to call the tasks cleaning up. However, I thought they might be implement such tasks for the future by the extension API and I'd like to keep it.

Even so should we remove this categorization?

nrc: I'm a bit uncomfortable about that - this is going to look like a mistake until/unless they add such an API, and when/if they do they might have a different name.

saneyuki: @nrc 

Okay. I'll remove do it.

I'd like to confirm the review to you. Is it what you expected change that to remove `cargo clean` task?

nrc: No, just to remove the `group` field (I think `group` is optional, right?).

saneyuki: > No, just to remove the group field.

Okay. Thank you.

> I think group is optional, right?

Yep.

nrc:
Thanks!

PR #135, opened 2017-08-25
proposal: Add 'cargo check' to the builtin task set.

saneyuki:


nrc:
There doesn't seem much point in using `cargo check` since this is essentially what the RLS is running when it is checking the files. The motivation for `cargo build`, etc., is that there are outputs which the RLS cannot produce. What is the motivation here?

saneyuki:
IMO, a language server sometimes behaves flaky because they have tables for source code and update/refresh incrementally by the bug. e.g. vscode's tsserver sometimes misses to list up errors in other files caused by editing the source code.  Of course, this is a implementation problem. But I seem it's useful to rechecking from zero for robustness. Furthermore, with a user hat, I'd like to avoid this problem with calling `cargo check` from a command parrete until the distributed `rls` binary has been fixed such the bug.

I recognized that this pull request is too careless. To resolve the above problem, it might good to introduce "trash the analysis result and re-analyse the project" command to rls. Or "warmup'' subcommand to analyse codes under the current directory.

@nrc How about this?

nrc:
> trash the analysis result and re-analyse the project

This is a good idea. The `cargo clean` command does this to some extent because we watch the target directory (with the RLS that just landed upstream, so maybe only from today's nightly). But I think it would be useful to have a more fine-grained command which:

* doesn't delete deps
* cleans the RLS's cache of analysis (though we pretty much do this on every build)
* maybe tries to rebuild the VFS (not sure if VSCode will let us do that, we'd need all changes to all changed files)
* starts a Cargo build (rather than just a rustc build)

This would be both quicker (because it doesn't clean the deps) and more thorough (because it resets the VFS) than doing `cargo clean`.

Much more simply, we might alias `cargo clean` to a `reset` command to make clear that it can be used to 'start over' the RLS.

> Or "warmup'' subcommand to analyse codes under the current directory.

This doesn't yet make sense for the RLS because we don't do any incremental checking.

saneyuki:
@nrc 

I'm sorry my reply was too late.

By your comment, I think it's better to close this pull request, and it's more proper way to work in rls side. Thank you for your advice!

PR #137, opened 2017-08-27
Prevent error when rustc is not in $PATH

killercup:
...by extending `$PATH` with a common place where rustup drops binaries.

Story time! I keep having this weird issue when opening a Rust project in vscode:

```
TypeError: Cannot read property 'toString' of null
	at makeRlsEnv (/Users/pascal/.vscode-insiders/extensions/rust-lang.rust-0.2.2/out/src/extension.js:44:36)
	at makeRlsProcess (/Users/pascal/.vscode-insiders/extensions/rust-lang.rust-0.2.2/out/src/extension.js:55:17)
	at autoUpdate.then (/Users/pascal/.vscode-insiders/extensions/rust-lang.rust-0.2.2/out/src/extension.js:114:57)
	at <anonymous>
```

Debugging the extension lead me to `src/extension.ts:41` [1], which assumes that the return value of `spawnSync` has a `stdout` property that can be converted to a string. Sadly, this is not always the case: When `spawn` could not find the binary to execute, it returns an object with an `error` property and also `stdout: null`.

I have no clue why it can't find `rustc` on my machine, but I sure have a weird setup. I'm using macOS, my default shell is fish, and this is in version `1.16.0-insider` of vscode. I'm using rustup, and all rust-related binaries are only available locally in `~/.cargo/bin/` from what I can tell.

This change should not cause any trouble---but I have only tested it on macOS. I'm not sure how other system (Windows) handles `$PATH` or even what syntax it expects. If vscode-rust targets Windows, patches to this may be required.

[1]: https://github.com/rust-lang-nursery/rls-vscode/blob/63000e65870f18318615c27a72c683cec05a3cbf/src/extension.ts#L41

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -25,27 +25,53 @@ import * as is from 'vscode-languageclient/lib/utils/is';
 
 const CONFIGURATION = RLSConfiguration.loadFromWorkspace();
 
+function getSysroot(env: Object): string | Error {
+    const rustcSysroot = child_process.spawnSync('rustc', ['--print', 'sysroot'], { env });
+
+    if (rustcSysroot.error) {
+        return new Error(`Error running \`rustc\`: ${rustcSysroot.error}`);
+    }
+
+    if (rustcSysroot.status > 0) {
+        return new Error(`Error getting sysroot from \`rustc\`: exited with \`${rustcSysroot.status}\``);
+    }
+    
+    if (!rustcSysroot.stdout || typeof rustcSysroot.stdout.toString !== 'function') {
+        return new Error(`Couldn't get sysroot from \`rustc\`: Got no ouput`);
+    }
+
+    const sysroot = rustcSysroot.stdout.toString()
+        .replace('\n', '').replace('\r', '');
+
+    return sysroot;
+}
+
 // Make an evironment to run the RLS.
 // Tries to synthesise RUST_SRC_PATH for Racer, if one is not already set.
 function makeRlsEnv(): any {
-    let env = process.env;
-    if (typeof env.HOME === 'string' && env.HOME.length) {
-        env.PATH = `${env.HOME}/.cargo/bin:${env.PATH || ""}`;
-    }
+    const env = process.env;
+
     if (process.env.RUST_SRC_PATH) {
         return env;
-    } else {
-        // rustc --print sysroot + lib/rustlib/src/rust/src
-        const result = child_process.spawnSync('rustc', ['--print', 'sysroot'], { env });
-        if (result.status > 0) {
-            console.log('error running rustc for sysroot');
-            return env;
-        }
-        let sysroot = result.stdout.toString();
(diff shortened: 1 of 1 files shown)


nrc:
> When spawn could not find the binary to execute, it returns an object with an error property and also stdout: null.

Could we handle this case gracefully by checking or catching the exception and warning the user?

Could we also only do the PATH hack if we fail to run rustup initially? It seems like this might actually break some users who have a working system with a weird PATH setup?

killercup:
Sure, this was just my quickest fix that worked. I'll try to work on that
in the morning.

Nick Cameron <notifications@github.com> schrieb am So. 27. Aug. 2017 um
23:36:

killercup:
Rebased, added a bunch of logging, and now, if there was an error, it prepends `~/.cargo/bin` to `$PATH`  and tries again.

src/extension.ts
```diff
@@ -25,24 +25,53 @@ import * as is from 'vscode-languageclient/lib/utils/is';
...
+
+    let result = getSysroot(env);
+    if (result instanceof Error) {
+        console.info(result.message);
+        console.info(`Let's retry with extended $PATH`);
+        env.PATH = `${env.HOME || '~'}/.cargo/bin:${env.PATH || ''}`;
+        result = getSysroot(env);
+
+        if (result instanceof Error) {
+            console.warn(result);
```
nrc: Could you use `window.showWarningMessage` here so the user will see it if they don't have their console open. Also, could you make a user-friendly message about not being able to set RUST_SRC_PATH because of an error reading the Rust sysroot?

src/extension.ts
```diff
@@ -25,24 +25,53 @@ import * as is from 'vscode-languageclient/lib/utils/is';
 
 const CONFIGURATION = RLSConfiguration.loadFromWorkspace();
 
+function getSysroot(env: Object): string | Error {
+    const rustcSysroot = child_process.spawnSync('rustc', ['--print', 'sysroot'], { env });
```
killercup: Now that I see this, I'm wondering if this should be `rustup run nightly rustc --print sysroot`. I think I saw RLS being called like that, and it'd probably be weird for racer to use another sysroot than RLS.

Do you want me to change this, @nrc?

src/extension.ts
killercup: Sure, will push in a second. I didn't want to introduce a new user-visible warning, but f you want one, I'll be happy to make one :)

src/extension.ts
nrc: Hmm, yeah, good catch, sounds like we should do that here too

killercup: Updated to use nighty, and also changed the cargo build call a bit further down (that get triggered when rls_root is set) to explicitly build using nightly, as RL will always be run on nightly IIUC.

nrc:
Thank you!

PR #138, opened 2017-08-27
Use better Rust logo for marketplace

nelsonjchen:
Instead of the existing logo, use the official colored 128x128 pixel Rust logo from:

https://www.rust-lang.org/logos/rust-logo-128x128.png

* Removes excessive negative space by being edge-to-edge. Previous logo looked small and diminutive.
* Colored, so it'll contrast much better on dark background themes. Previous logo was black on grey on the _default_ VSCode theme.

- - -
_After examples are simulated by simply replacing the distributed package's `rust-icon.png` and reloading._

Before:
screen shot 2017-08-27 at 11 00 17 am

After:
screen shot 2017-08-27 at 11 07 03 am

After(Light):
screen shot 2017-08-27 at 11 08 04 am

Negative Space Comparison:
less negative space


nrc:
Thanks! This looks much nicer!

nelsonjchen:
Indeed it does. Though, I would prefer a higher contrast and brighter orange color version like the one on my socks but that doesn't seem to be immediately available. This will do.

Issue #100, opened 2017-07-29
Feature: Implement trait for struct


doxxx:
It would be really useful to be able to generate the function skeletons necessary to implement a trait for a struct. So for example, if I have the following:

```rust
trait Foo {
    fn do_stuff(s: &str) -> u64;
}

struct Bar {
    x: u64,
}

impl Foo for Bar {
//cursor
}
```

With the cursor at the indicated location, invoking the `Implement Trait for Struct` command would generate the following:

```rust
impl Foo for Bar {
    fn do_stuff(s: &str) -> u64 {
        unimplemented!();
    }
}
```

nrc:
This would be great!

Some thoughts on doing this:
* afaict, there is not a good way to give input to vscode for things like refactoring, so this would probably have to work by typing `impl Foo for Bar` and then expanding that (I guess we could just type and select `Foo` and expand from there too) 
* we should probably handle super-traits - save-analysis knows about sub/super trait relationships (or did, might have been removed actually), but the RLS does not. I don't think this is that hard.
* We already know all the items in a trait, so I think we just have to copy their signatures, and replace `;` with `{ unimplemented! }` for functions
* However, we don't include the signature info for the RLS to save space. We would need to be able to fetch this from the compiler on demand (including for previously compiled crates, which would be challenging).
* I wonder what we should do for default methods? Provide stubs or not?
* Should we replace `Self` by `Bar`? We don't need to, I'm not sure which is more idiomatic.
* Are there any other type substitutions we need to do? (I don't think so)
* Ideally we would do any imports that need doing, but we don't have a mechanism for that yet.

doxxx:
If I understand you correctly, RLS doesn't currently store function signatures which you would need for this. They would probably also be helpful for another feature I'd love to see: function parameter info popups.

nrc:
> If I understand you correctly, RLS doesn't currently store function signatures

Correct. Save-analysis can provide it (the new rustdoc will use it) but the RLS requests not to get it to save space in the data (one of the biggest sources slowdown is processing large chunks of JSON data, so this is an important optimisation). We might be able to use a slimmed down version for both functon signature assistance (which I'd also love to have) and for this feature.

doxxx:
[...]

PR #141, opened 2017-08-29
snippet: Add many usefull snippets

Kerollmops:
These snippets comes from https://github.com/editor-rs/vscode-rust

### DIFF
--- a/snippets/rust.json
+++ b/snippets/rust.json
@@ -340,8 +340,8 @@
     "extern-fn": {
         "prefix": "extern-fn",
         "body": [
-            "extern \"C\" fn ${1:name}(${2:arg}: ${3:Type}) -> ${4:RetType} {",
-            "    ${5:// add code here}",
+            "extern \"C\" fn ${1:name}(${2}) ${3}{",
+            "    ${4:// add code here}",
             "}"
         ],
         "description": "extern \"C\" fn …(…) { … }"
--- a/snippets/rust.json
+++ b/snippets/rust.json
@@ -458,9 +458,8 @@
     "match": {
         "prefix": "match",
         "body": [
-            "match ${1:expr} {",
-            "    ${2:Some(expr)} => ${3:expr},",
-            "    ${4:None} => ${5:expr},",
+            "match ${1} {",
+            "    ${2} => ${3},",
             "}"
         ],
         "description": "match … { … }"
--- a/snippets/rust.json
+++ b/snippets/rust.json
@@ -506,7 +506,7 @@
         "body": [
             "#[derive(Debug)]",
             "struct ${1:Name} {",
-            "    ${2:field}: ${3:Type}",
+            "    ${2}: ${3:Type}",
             "}"
         ],
         "description": "struct … { … }"
--- a/snippets/rust.json
+++ b/snippets/rust.json
@@ -358,8 +358,8 @@
     "fn": {
         "prefix": "fn",
         "body": [
-            "fn ${1:name}(${2:arg}: ${3:Type}) -> ${4:RetType} {",
-            "    ${5:unimplemented!();}",
+            "fn ${1:name}(${2}) ${3}{",
+            "    ${4:unimplemented!();}",
             "}"
         ],
         "description": "fn …(…) { … }"
--- a/snippets/rust.json
+++ b/snippets/rust.json
@@ -1,26 +1,199 @@
 {
-    "for": {
-        "prefix": "for",
+    "allow": {
+        "prefix": "allow",
         "body": [
-            "for $1 in $2 {",
-            "\t$0",
-            "}"
+            "#![allow(${1})]"
         ],
-        "description": "Insert for loop"
+        "description": "#![allow(…)]"
     },
-    "unimplemented": {
-        "prefix": "unimplemented",
+    "deny": {
+        "prefix": "deny",
         "body": [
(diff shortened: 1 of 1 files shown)


nrc:
Thanks for the PR! I would like to extend the snippets we have available, but I'm a bit cynical about adding every possible snippet en mass (and I'm not sure that taking these snippets as-is from vscode-rust without checking they are optimal). Are there any snippets you particularly miss having? Or that you feel would be high impact to have here?

Kerollmops:
Ahah yes I'm missing `impl Trait for Xxx`, `fn`, `struct`, `struct ()`, `struct ;` and probably others.

I looked at the `impl Trait for Xxx` that can be dynamic, vscode doesn't seem to handle this for the moment https://github.com/Microsoft/vscode/issues/6920, https://github.com/rust-lang-nursery/rls-vscode/issues/100.

Object905:
Snippets for attributes with prefix without `#![` can be really annoying, because when you want to write some attribute for example `#![feature(..)]` you start by writing `#![`, then `featu...` and snippet with prefix `feature` will show up and if you choose it you end up with something like `#![#![feature(..)]`. Maybe change prefix of such snippets to `#![keyword`. And it also would be useful to duplicate for `#[keyword` for attributes on a single item.

snippets/rust.json
```diff
@@ -40,6 +453,117 @@
...
+            "    ${1:unimplemented!();}",
+            "}"
+        ],
+        "description": "fn main() { … }"
+    },
+    "match": {
+        "prefix": "match",
+        "body": [
+            "match ${1:expr} {",
+            "    ${2:Some(expr)} => ${3:expr},",
```
suhr: Should be just `"    $2"`

snippets/rust.json
```diff
@@ -40,6 +453,117 @@
...
+            "struct ${1:Name};"
+        ],
+        "description": "struct …;"
+    },
+    "struct": {
+        "prefix": "struct",
+        "body": [
+            "#[derive(Debug)]",
+            "struct ${1:Name} {",
+            "    ${2:field}: ${3:Type}",
```
suhr: Should be just `"    $2"`

snippets/rust.json
```diff
@@ -1,33 +1,446 @@
...
+        "prefix": "extern-crate",
+        "body": [
+            "extern crate ${1:name};"
+        ],
+        "description": "extern crate …;"
+    },
+    "extern-fn": {
+        "prefix": "extern-fn",
+        "body": [
+            "extern \"C\" fn ${1:name}(${2:arg}: ${3:Type}) -> ${4:RetType} {",
```
suhr: Should be `"extern \"C\" fn ${1:name}($2) $3 {"`

snippets/rust.json
```diff
@@ -1,33 +1,446 @@
...
+            "extern \"C\" {",
+            "    ${2:// add code here}",
+            "}"
+        ],
+        "description": "extern \"C\" { … }"
+    },
+    "fn": {
+        "prefix": "fn",
+        "body": [
+            "fn ${1:name}(${2:arg}: ${3:Type}) -> ${4:RetType} {",
```
suhr: Should be `"fn ${1:name}($2) $3 {`

snippets/rust.json
```diff
@@ -1,33 +1,446 @@
...
+        "description": "stringify!(…)"
+    },
+    "thread_local": {
+        "prefix": "thread_local",
+        "body": [
+            "thread_local!(static ${1:STATIC}: ${2:Type} = ${4:init});"
+        ],
+        "description": "thread_local!(static …: … = …);"
+    },
+    "try": {
```
suhr: Do we still need `try!`?

snippets/rust.json
```diff
@@ -1,33 +1,446 @@
...
+            "}"
+        ],
+        "description": "else { … }"
+    },
+    "enum": {
+        "prefix": "enum",
+        "body": [
+            "#[derive(Debug)]",
+            "enum ${1:Name} {",
+            "    ${2:Variant1},",
```
suhr: Should be just `"    $2"`

snippets/rust.json
Kerollmops: nope

snippets/rust.json
Kerollmops: IFAIK the "Variant1" text is just the default text placed, no ?

snippets/rust.json
Kerollmops: even the `RetType` doesn't need a default name

Kerollmops: Ho yes sorry, I didn't understand your change proposition

snippets/rust.json
Kerollmops: yes but in some cases it's just a hint for the user

Kerollmops: and you have just to write hover it

Kerollmops:
Up on this PR ?

nrc:
Sorry, perhaps I was unclear with my earlier comment. This PR seems to add every snippet from vscode-rust, I don't want to do that. I am not convinced there is value in it (and there are downsides in having so many snippets - it can be confusing for users and contributors, and it can make code completion worse).

So, I would rather add a few snippets at a time ensuring that each snippet is well-motivated and optimally specified.

PR #148, opened 2017-09-18
update dependencies; should fix travis

genneth:
Maybe a bit aggressive on pushing deps forwards? Not sure what the right balance is...


nrc:
Thank you!

PR #150, opened 2017-09-19
Be able to restart the RLS server

### LINKED ISSUE #129: Add command to restart the RLS
I can't find docs on how to do this exactly, but I expect one sends the `shutdown` and `exit` messages, then calls `stop`, then `start` on the language client. (But perhaps `stop` sends the LSP messages?).

genneth:
Tries to fix #129. Whilst this works, I'm not very happy with the structure. I think overall, we need a bit more scaffolding around separately storing things necessary for the extension, the workspace, etc. I've been looking at the source for vscode-powershell for inspiration.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -173,19 +173,7 @@ function startLanguageClient(context: ExtensionContext)
     lc = new LanguageClient('Rust Language Server', serverOptions, clientOptions);
     lcOutputChannel = lc.outputChannel;
 
-    let runningDiagnostics = 0;
-    lc.onReady().then(() => {
-        lc.onNotification(new NotificationType('rustDocument/beginBuild'), function(_f) {
-            runningDiagnostics++;
-            startSpinner('RLS: working');
-        });
-        lc.onNotification(new NotificationType('rustDocument/diagnosticsEnd'), function(_f) {
-            runningDiagnostics--;
-            if (runningDiagnostics <= 0) {
-                stopSpinner('RLS: done');
-            }
-        });
-    });
+    diagnosticCounter();
 
     const disposable = lc.start();
     context.subscriptions.push(disposable);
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -29,41 +28,41 @@ function getSysroot(env: Object): string | Error {
     const rustcSysroot = child_process.spawnSync(
         'rustup', ['run', 'nightly', 'rustc', '--print', 'sysroot'], {env}
     );
-
+
```
nrc: Could you avoid inserting all this extra whitespace please

src/extension.ts
```diff
@@ -29,41 +28,41 @@ function getSysroot(env: Object): string | Error {
...
+    
     if (!rustcSysroot.stdout || typeof rustcSysroot.stdout.toString !== 'function') {
         return new Error(`Couldn't get sysroot from \`rustc\`: Got no ouput`);
     }
-
+    
     const sysroot = rustcSysroot.stdout.toString()
-        .replace('\n', '').replace('\r', '');
-
+    .replace('\n', '').replace('\r', '');
```
nrc: or removing it :-)

src/extension.ts
```diff
@@ -159,22 +167,32 @@ export function activate(context: ExtensionContext) {
...
-    const lc = new LanguageClient('Rust Language Server', serverOptions, clientOptions);
+    lc = new LanguageClient('Rust Language Server', serverOptions, clientOptions);
     lcOutputChannel = lc.outputChannel;
-
-    diagnosticCounter(lc);
-    registerCommands(lc, context);
-    activateTaskProvider();
-
+    
+    let runningDiagnostics = 0;
```
nrc: Why do you want to inline this code?

genneth:
I tried to look for an unregister, but couldn't find it. I suspect it doesn't exist. The vscode-powershell essentially has global variables: it organises them carefully, inside an overall `SessionManager` object, the constructor of which does things like registering commands, and separately `start` and `stop` which can be called multiple times.

src/extension.ts
genneth: Sure; a lot of this was over zealously letting VSCode decide on formatting.

src/extension.ts
genneth: It didn't seem to be obviously a reusable function --- calling it repeatedly doesn't make sense?

nrc:
I filed #152 for the refactoring

src/extension.ts
nrc: It's more for self-documentation - I found it made the `activate` function easy to read by factoring out self-contained bits of functionality. Inlining them made for a long, hard to grok function.

genneth:
Tweaked the patch to be minimal delta against master

nrc:
Thank you!

PR #154, opened 2017-09-21
Set LD_LIBRARY_PATH even when RUST_SRC_PATH is already set. (Don't return early.)

golddranks:
Great timing with this commit! https://github.com/rust-lang-nursery/rls-vscode/commit/14f32304d99885fd2aef8e69d1c20754766ab328 Turns out my RLS would just crash without it, since it load wrong libraries.

However, I had the `RUST_SRC_PATH` environment variable set in my environment, which made `makeRlsEnv` return early and not set `RUST_SRC_PATH`, so I changed that.


nrc:
Thanks!

PR #156, opened 2017-09-21
Add onEnter rules for comments

### LINKED ISSUE #153: Continue Comment on New Line
Other language extensions in VS Code (such as Python) continue comments on a new line when pressing `enter`. This makes typing a little more fluid, especially when writing long doc comments. I would be happy to work on this; I will work on the PR if there is consensus that this is a good idea.

Current functionality:

```rust
/// Tests a scenario<newline>
<-- Cursor here
fn test() {
}
```

Proposed functionality:

```rust
/// Tests a scenario<newline>
/// <-- Cursor here
fn test() {
}
```

DSpeckhals:
This adds the convenience of VSCode's hooks for onEnter events for Rust. Now, when pressing `enter` within a  comment, the comment continues. For example: after typing `///` and pressing `enter`, the next line that the cursor is on will also have `///`, followed by a single space.

`//!`, `/**`, and `/*!` multi-line comments are also supported. Single line comments (`//`), are not continued on `enter`.

This fixes #153.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -254,10 +254,16 @@ function configureLanguage(context: ExtensionContext) {
     const disposable = languages.setLanguageConfiguration('rust', {
         onEnterRules: [
             {
-                // Begins a triple-slash doc comment
-                // e.g. ///| or //!|
-                beforeText: /^\s*\/{2}(\/|\!).*$/,
+                // Doc single-line comment
+                // e.g. ///|
+                beforeText: /^\s*\/{3}.*$/,
                 action: { indentAction: IndentAction.None, appendText: '/// ' },
+            },
+            {
+                // Parent doc single-line comment
+                // e.g. //!|
+                beforeText: /^\s*\/{2}\!.*$/,
+                action: { indentAction: IndentAction.None, appendText: '//! ' },
             },
             {
                 // Begins an auto-closed multi-line comment (standard or parent doc)
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -18,7 +18,8 @@ import { activateTaskProvider, deactivateTaskProvider } from './tasks';
 import * as child_process from 'child_process';
 import * as fs from 'fs';
 
-import { workspace, ExtensionContext, TextEditor, TextEditorEdit, window, commands, OutputChannel } from 'vscode';
+import { commands, ExtensionContext, IndentAction, languages, OutputChannel, TextEditor,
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -247,3 +249,41 @@ function registerCommands(context: ExtensionContext) {
     });
     context.subscriptions.push(restartServer);
 }
+
+function configureLanguage() {
+    languages.setLanguageConfiguration('rust', {
```
Xanewok: Since `setLanguageConfiguration()` returns an `Disposable`, I think we should to register it as a disposable, e.g. what [official typescript extension](https://github.com/Microsoft/vscode/blob/master/extensions/typescript/src/typescriptMain.ts#L136) does.

DSpeckhals: Good call. I fixed it.

src/extension.ts
```diff
@@ -247,3 +249,42 @@ function registerCommands(context: ExtensionContext) {
...
 }
+
+function configureLanguage(context: ExtensionContext) {
+    const disposable = languages.setLanguageConfiguration('rust', {
+        onEnterRules: [
+            {
+                // Begins a triple-slash doc comment
+                // e.g. ///| or //!|
+                beforeText: /^\s*\/{2}(\/|\!).*$/,
+                action: { indentAction: IndentAction.None, appendText: '/// ' },
```
Xanewok: Currently this always inserts `/// `, even if in a module comment `//! `.
Would it be possible to specify separate rules for `///` and `//!` respectively? Didn't toy with that myself, so I have no idea if that's easy/possible at all.

Xanewok:
Looks good, thanks for the `//!` fix!

src/extension.ts
DSpeckhals: You're right. Fixed. Thanks for the catch

nrc:
Thanks for the PR! Could you try doing this in language-configuration.json, rather than programmatically? I think everything can be done that way.

DSpeckhals:
@nrc Right. That's what I was thinking originally. I tried it earlier, and VSCode doesn't support extensions loading `onEnterRules` from JSON at them moment. Sadly, this configuration has to be set in code.

See https://github.com/Microsoft/vscode/issues/11514 for more context.

nrc:
Ah, ok, thanks for the explanation!

fatihaziz:
+1
5 years later I dont know why in javascript it doesn't work....

PR #157, opened 2017-09-21
Display warning message if Cargo.toml is missing

### LINKED ISSUE #75: detect a rust file outside a Cargo project and inform user
I.e., opening a rust file without opening the whole project as a folder. This makes the RLS unhappy. We should detect this situation and not start the RLS (or let it exit gracefully) and inform the user.

DSpeckhals:
If there is no Cargo.toml at the root of the workspace, display a warning message to the user informing them that not all features will be available.

This fixes #75

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -155,6 +155,8 @@ export function activate(context: ExtensionContext) {
 
 function startLanguageClient(context: ExtensionContext)
 {
+    warnOnMissingCargoToml();
+
     window.setStatusBarMessage('RLS: starting up');
 
     // FIXME(#66): Hack around stderr not being output to the window if ServerOptions is a function
@@ -190,6 +192,14 @@ export function deactivate(): Promise<void> {
     deactivateTaskProvider();
 
     return Promise.resolve();
+}
+
+function warnOnMissingCargoToml() {
+    workspace.findFiles('Cargo.toml').then(files => {
+        if (files.length < 1) {
+            window.showWarningMessage('Cargo.toml must be in the workspace in order to support all features');
+        }
+    });
 }
 
 function warnOnRlsToml() {


Xanewok:
Maybe it'd be good to use the [`"workspaceContains":Cargo.toml`](https://code.visualstudio.com/docs/extensionAPI/activation-events#_activationeventsworkspacecontains) rule in `package.json` instead? This looks like it will take care of appropriately calling the `activate()`, registering commands when needed etc.
For example `omnisharp-vscode` does [that](https://github.com/OmniSharp/omnisharp-vscode/blob/master/package.json#L206).

DSpeckhals:
@Xanewok Unless I'm reading this wrong, the package wouldn't activate at all in that case. This includes the RLS, but also the syntax features and rls.update command. I guess it's not a big deal. I'm good either way on this one.

Xanewok:
@DSpeckhals ah, my bad. It seems the activation events will be sent when something happens
(e.g. when vscode is initialized in a workspace folder with _Cargo.toml_ if there's a  `"workspaceContains":Cargo.toml` option), but that only lazily starts the extension automatically and is not a required condition for the extension to operate.

Coincidentally, I added a PR https://github.com/rust-lang-nursery/rls/pull/492 that prevents the RLS from crashing and dying when the underlying Cargo routine fails and allows it to be successfully re-run, without having to restart the RLS.
Right now I tried creating an empty folder with a _main.rs_, initialized rls there and only after, inside vscode, I added the _Cargo.toml_ and moved _main.rs_ under _src/_ and it seemed to be able to work properly!

With that in mind, I don't think we have to go as far as not running the RLS at all if the workspace does not have a Cargo.toml, but it would be really useful to report that fact to the user, either via a status bar or a single warning like `Cargo.toml has to be present in in the workspace in order to support all the features` or something along the lines.

DSpeckhals:
That sounds good to me. We can check for the existence of the Cargo file
then, asynchronous from the test of the activation.

I can modify this PR to do what you describe about the message instead, and
forget about the condition for starting the RLS.

Xanewok:
Yeah, I think only presenting the warning on startup is enough now :smiley:
Thanks!

cc @nrc

DSpeckhals:
I force-pushed a commit on top of my previous one, changed the PR description, and changed the PR title to reflect the current state.

nrc:
Thank you!

maelvalais:
Note: it would be good to change it from

    Cargo.toml must be in the workspace in order to support all features

to 

    A Cargo.toml file must be at the workspace root in order to support all features

No?

Xanewok:
@maelvalais Yup! Especially since workspace mode should work better now. Mind sending a PR with the change?

PR #158, opened 2017-09-21
Deprecate `rust-client.showStdErr` property.

### LINKED ISSUE #149: rust-client.showStdErr is not respected
Updating deps in https://github.com/rust-lang-nursery/rls-vscode/pull/148 pulled a fix https://github.com/Microsoft/vscode-languageserver-node/commit/8ed63637d95e1dbd7ad0b6e5d71195a7a88943b2, which now always logs the stderr of the language server to the related output window in the VSCode.

@nrc do you think we could remove the `rust-client.showStdErr` option and the related [hack](https://github.com/rust-lang-nursery/rls-vscode/blob/master/src/extension.ts#L144) now?

Xanewok:
Closes #149.

#148 pulled in a 3.4.2 version of vscode-languageclient, which fixes the bug (#66) with stderr not being properly appended if the server options were a function.
With #148, the property doesn't change the way the extension behaves anymore - it logs the stderr, no matter if the `rust-client.showStdErr` is set or not. The changes here mostly prune what's not required anymore.

Not sure how we should proceed with the deprecated property. Should it stay marked as `@deprecated` in the API? Should we just flat-out remove it and not care? When vscode encounters a property it doesn't recognize, it only highlights it with a warning saying `Unknown configuration setting`, so just removing the property would not break anything for the users.

### DIFF
--- a/package.json
+++ b/package.json
@@ -142,11 +142,6 @@
             "type": "object",
             "title": "Rust configuration",
             "properties": {
-                "rust-client.showStdErr": {
-                    "type": "boolean",
-                    "default": false,
-                    "description": "Specifies whether RLS stderr will be shown in the output channel along with protocol logs. Requires reloading extension after change."
-                },
                 "rust-client.logToFile": {
                     "type": "boolean",
                     "default": false,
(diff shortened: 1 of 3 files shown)


nrc:
I would just rip it out - we're pre-1.0 and it only makes sense to use it for debugging, so I think it is fine to just remove.

Xanewok:
Done, removed the property and force-pushed.

nrc:
Thanks!

PR #161, opened 2017-09-29
Added custom path for rustup

### LINKED ISSUE #160: Custom rustup location
Hi.

In Windows 10 there is a subsystem called [WSL][1]. 
I want to use `rust` installed on Linux, while still working on Windows.

To do so I need to be able to pass custom `rustup` path to `rls-vscode` .

[1]: https://en.wikipedia.org/wiki/Windows_Subsystem_for_Linux

yisonPylkita:
Closes #160

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -28,6 +28,7 @@ function fromStringToRevealOutputChannelOn(value: string): RevealOutputChannelOn
 }
 
 export class RLSConfiguration {
+    public readonly rustupPath: string;
     public readonly logToFile: boolean;
     public readonly revealOutputChannelOn: RevealOutputChannelOn = RevealOutputChannelOn.Never;
     public readonly updateOnStartup: boolean;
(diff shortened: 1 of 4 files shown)


src/configuration.ts
```diff
@@ -44,14 +45,15 @@ export class RLSConfiguration {
      * directory.
      */
     public readonly rlsRoot: string | null;
-
+
```
nrc: nit: please don't add whitespace

src/configuration.ts
```diff
@@ -44,14 +45,15 @@ export class RLSConfiguration {
...
-
+    
     public static loadFromWorkspace(): RLSConfiguration {
         const configuration = workspace.getConfiguration();
 
         return new RLSConfiguration(configuration);
     }
 
     private constructor(configuration: WorkspaceConfiguration) {
+        this.rustupPath = configuration.get('rustup.path', 'rustup');
```
nrc: Could you use something like`rust-client.rustupPath` for the configuration name please? Could you also add the option to the `configuration` section of package.json.

src/configuration.ts
yisonPylkita: Ok, I will correct this and pay attention to this in future.

src/configuration.ts
yisonPylkita: Yeah, you are right.

I will submit patch when I get home.

nrc:
Looks good, thanks! Could you rebase please?

yisonPylkita:
Sorry for the delay.

I have updated this branch to an actual master. Is this a proper way of doing this ?

yisonPylkita:
Sorry for the delay. 
I have rebased this branch onto actual master.

Xanewok:
Thank you!

PR #169, opened 2017-10-03
Check for new  `rls-preview` component when detecting RLS components

Xanewok:
Follows https://github.com/rust-lang-nursery/rls-vscode/commit/ffd9ca4616902850019bd0f07b3e55c1970efeef.
Fixes detecting installed `rls-preview` instead of `rls` (won't say that it's not installed when the user has in fact installed `rls-preview`.

@nrc should we release 0.3.1 with this fix?


nrc:
Thanks, yeah, I'll do a release soon.

PR #173, opened 2017-10-13
Remove dependency on nightly

ykomatsu:
rls-vscode should work with beta and stable channel.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -27,7 +27,7 @@ export const CONFIGURATION = RLSConfiguration.loadFromWorkspace();
 
 function getSysroot(env: Object): string | Error {
     const rustcSysroot = child_process.spawnSync(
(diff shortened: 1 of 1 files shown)


raggy-rs:
Yes please!
Without this change the extension is unusable even with nightly if you have nightly-x86_64-pc-windows-msvc installed because nightly defaults to the nightly-i686-pc-windows-gnu.

ghisvail:
You might also want to update the README for completeness:
```
## Requirements
[...]
* A nightly Rust toolchain (the extension will configure this for you, with
  permission)
```
to
```
* A Rust toolchain (the extension will configure this for you, with
  permission)
```

Xanewok:
Yeah, we're using `CONFIGURATION.channel` everywhere else, and since it's available on every channel now, it seems like a good idea to unify this. Thanks!

ghisvail:
Could you release `v0.3.2` with this change in please?

PR #178, opened 2017-10-26
Remove extension-side problem matchers to remove duplicate error/warnings

### LINKED ISSUE #133: Duplicate output in the "Problems" panel
untitled

jTitor:
(fixes #133)

The language server now directly posts diagnostics via LSP, so the extension doesn't have to do parsing anymore; this was causing #133.

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,39 +105,6 @@
                 }
             ]
         },
-        "problemMatchers": [
-            {
-                "name": "rustc",
-                "owner": "rust",
-                "fileLocation": [
-                    "relative",
-                    "${workspaceRoot}"
-                ],
-                "pattern": [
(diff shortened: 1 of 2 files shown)


nrc:
I don't think this is the correct fix for #133 - Although the RLS does supply some errors of its own, these problem matchers are for when the user runs the `cargo build` (or similiar) tasks. If the user does that, then it is nice to have these matchers.

Xanewok:
Agreed, it'd be good to keep them to be honest. It might be worth following [vscode-go](https://github.com/Microsoft/vscode-go/pull/2109) and deduplicating the diagnostics on the extension side. I'll close the PR but thank you regardless!

Issue #179, opened 2017-10-26
Choose a channel for the RLS based on the current Rustup default.


nrc:
E.g., if the user is using the stable channel by default, then we should use the stable channel for launching the RLS with Rustup.

This should not be used, if the user has set `rust-client.channel`, which means we will need to change the default for that setting to `null` rather than `"nightly"`.

To get started, look at rustup.ts. I think you'll need to parse the result of `rustup show`.

lnicola:
I wonder if there shouldn't be a `rustup run default foo` that uses the default channel instead.

lnicola:
@nrc is there any reason to invoke `rls` through `rustup` when the channel hasn't been customized?

nrc:
Yeah, that might work. I can't think of a reason why it would not. We'd need to experiment to make sure.

We'd need to use rustup if just calling `rls` didn't work (to install RLS or whatever), but then I think we could just use rustup without specifying a channel to get the default.

lnicola:
> I can't think of a reason why it would not.

Maybe `rustup` is in `PATH`, but `~/.cargo/bin` isn't.

> I think we could just use rustup without specifying a channel to get the default.

The thing is that `rustup run` seems to require a channel name.

lnicola:
I opened https://github.com/rust-lang-nursery/rustup.rs/pull/1279, maybe we'll get this.

ctaggart:
Looks like the latest nightly for Mac is missing `rls-preview`, so I just had to install the one from yesterday:
```
rustup install nightly-2018-09-22
rustup default nightly-2018-09-22
rustup component add rls-preview rust-analysis rust-src
```

PR #180, opened 2017-10-27
Use the user's default toolchain

lvillani:
Since Rust 1.21 the RLS can be installed from the stable channel.

Switching to use the stable channel by default would let people who
prefer to depend only on the stable compiler to have nightly installed
for one less thing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -170,7 +170,7 @@
                 },
                 "rust-client.channel": {
                     "type": "string",
-                    "default": "nightly",
+                    "default": "default",
                     "description": "Rust channel to install RLS from."
                 },
                 "rust-client.rls-name": {
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -10,6 +10,8 @@
 
 'use strict';
 
+import { execSync } from 'child_process';
+
 import { workspace, WorkspaceConfiguration } from 'vscode';
 import { RevealOutputChannelOn } from 'vscode-languageclient';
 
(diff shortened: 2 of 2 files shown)


nrc:
In https://github.com/rust-lang-nursery/rls-vscode/issues/179 I propose a slightly different approach - using the current Rustup default as the default. So if the user is using stable, they'll get stable and if they're using nightly they'll get nightly. I'd prefer not to always use stable as default since for users with both nightly and stable toolchains, they'll get a significantly better experience with nightly.

lvillani:
Ah! That makes total sense. I'll try to update this PR later this week.

lvillani:
I changed the PR to look up for the default toolchain. Not sure how it fits with https://github.com/rust-lang-nursery/rustup.rs/pull/1279#issuecomment-340614615, though. Also not sure why the build on Travis is failing, it builds just fine for me.

package.json
```diff
@@ -170,7 +170,7 @@
                 },
                 "rust-client.channel": {
                     "type": "string",
-                    "default": "nightly",
+                    "default": "default",
```
nrc: I think you can use `null` rather than `"default"`

src/configuration.ts
```diff
@@ -58,15 +60,37 @@ export class RLSConfiguration {
...
-        this.channel = configuration.get('rust-client.channel', 'nightly');
-        this.componentName = configuration.get('rust-client.rls-name', 'rls');
+        this.channel = configuration.get('rust-client.channel', 'default');
+        if (this.channel === 'default') {
+            try {
+                this.channel = RLSConfiguration.defaultChannel(this.rustupPath);
+            } catch (e) {
+                console.log(e);
+            }
+        }
```
nrc: Could you move this code to rustup.rs?

nrc:
Not sure why Travis is failing. Looks like a TypeScript compilation error, but it's not in a place that I would have thought your code should affect.

src/configuration.ts
lvillani: I thought about moving it there but most of the code in `rustup.ts` relies on the `RLSConfiguration` object being already initialised, and is async. I thought this code would be inconsistent with the rest of the code there.

nrc: It looks like you're using `RLSConfiguration` here too though? I don't think that whether the code is sync or async should make too much difference as to where it should live.

nrc:
The test failure is weird - I see it locally too, but that code hasn't changed for months. I guess a library changed or something. Anyway, not your problem :-)

lvillani:
While working on this I found several edge cases and I have not enough time to handle them at the moment. I'm closing this PR for now.

PR #197, opened 2017-11-26
Fix typescript error _f implicitly has any type

bvinc:
```
src/extension.ts(207,85): error TS7006: Parameter '_f' implicitly has an 'any' type.
src/extension.ts(211,89): error TS7006: Parameter '_f' implicitly has an 'any' type.
```

I'm new to typescript, but from what I can tell, these errors are completely valid.  You set "strict: true" in tsconfig.json, which sets noImplicitAny.  I tried it with typescript 2.5.3 and 2.6.1 and always get these errors.  Why am I the only one getting these errors?

This PR fixes it.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -204,11 +204,11 @@ async function autoUpdate() {
 function diagnosticCounter() {
     let runningDiagnostics = 0;
     lc.onReady().then(() => {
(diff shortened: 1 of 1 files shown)


nrc:
Thanks!

PR #202, opened 2017-11-28
Improve snippet for `if let`

jonasbb:
I find the provided `if let` snippet is rather useless. Besides the obvious numbering error, I think offering `foo.` in the if block is not helpful. As such a simple if let snippet, which works will all enums and doesn't suggest anything inside the block would be more helpful.

### DIFF
--- a/snippets/rust.json
+++ b/snippets/rust.json
@@ -40,14 +40,14 @@
         ],
         "description": "Insert macro_rules!"
     },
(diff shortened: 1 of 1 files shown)


nrc:
I think it would be worth adding a separate `if let` snippet for the general case. I find that the existing snippet is useful for the equivalent of a null check and is a fairly common idiom.

PR #207, opened 2017-12-03
Tell the user that Cargo.toml should be found at the root

maelvalais:
This is a follow-up to the comments in #157
(https://github.com/rust-lang-nursery/rls-vscode/pull/157#issuecomment-348599867)

I changed the message

    Cargo.toml must be in the workspace in order to support all features

to 

    A Cargo.toml file must be at the root of the workspace in order to support all features

so that we are reminded that the Cargo.toml should be at the root


nrc:
Thanks!

PR #212, opened 2017-12-14
Do not start RLS if a file is opened instead of a folder

kichristensen:
RLS crashes if a file is opened instead of a folder.
This PR checks if that is the case, displays a warning message to the user and does not start the RLS.

RLS has at least two issues of this rust-lang-nursery/rls#628 and rust-lang-nursery/rls#619

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -143,6 +143,10 @@ export function activate(context: ExtensionContext) {
 
 function startLanguageClient(context: ExtensionContext)
 {
+    if (workspace.rootPath === undefined) {
+        window.showWarningMessage('Cannot start RLS because no folder was opened');
+        return;
+    }
     warnOnMissingCargoToml();
 
     window.setStatusBarMessage('RLS: starting up');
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -143,6 +143,10 @@ export function activate(context: ExtensionContext) {
 
 function startLanguageClient(context: ExtensionContext)
 {
+    if (workspace.rootPath === undefined) {
+        window.showWarningMessage('Cannot start RLS because no folder was opened');
+        return;
+    }
```
nrc: Could you factor this out into a function (like `warnOnMissingCargoToml`) please? For the warning message, I would prefer to negate the verb rather than the noun here, but it would be even better to phrase it positively, something like "Startup error: the RLS can only operate on a folder, not a single file"

kichristensen: @nrc sure, I'll make the changes. I just have one question, to avoid the error that RLS crashed multiple times and won't be started again, we have to return before trying to start RLS. If the logic is factored out into a function, how should this be handled?

nrc: ah, yeah, good point. I mean you could return a bool from the function, but I'm not sure that is worth doing. So maybe just leave it as is, but change the message.

kichristensen:
@nrc The message has now been changed as suggested

PR #214, opened 2017-12-20
add link to "Hello, Cargo!" from quick start

durka:
Many beginners seem to get tripped up when the Quick Start says "now open up your Rust project, i.e. the folder containing Cargo.toml" and they're like "I know some of those words". So this adds a link to the book section explaining how to create a new project.


nrc:
Thanks!

PR #218, opened 2017-12-28
feat(configuration): set default channel to stable

### LINKED ISSUE #217: Warn nightly users if rls is missing in the active toolchain
So I've been fighting for the past couple hours with `rls-preview` on the latest nightly build after doing an unwise `rustup update`. Turns out rls is missing in the latest nightly and has been missing on and off for the past few weeks due to breaking changes in nightly (as is to be expected). 

https://github.com/rust-lang-nursery/rls/issues/611

While trying to work out what was the last working nightly/rls release I saw it mentioned in the issue linked above that rls has a release on stable now (yay!), and sure enough setting  `rust-client.channel` to `stable` got rls working again with this extension. Might it make sense to default to stable rather than asking the user if they want to install the latest nightly, which has a small chance of missing `rls-preview`?

kwonoj:
- closes #217 

Feel freely close if this isn't desired direction -  as it's small changes I took stab directly via PR.
This PR changes default channel to stable, prevents rustup update nightly doesn't have rls-preview breaks extension behavior.

### DIFF
--- a/package.json
+++ b/package.json
@@ -165,7 +165,7 @@
                 },
                 "rust-client.channel": {
                     "type": "string",
-                    "default": "nightly",
+                    "default": "stable",
                     "description": "Rust channel to install RLS from."
                 },
                 "rust-client.rls-name": {
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -58,7 +58,7 @@ export class RLSConfiguration {
         this.revealOutputChannelOn = RLSConfiguration.readRevealOutputChannelOn(configuration);
         this.updateOnStartup = configuration.get<boolean>('rust-client.updateOnStartup', true);
 
-        this.channel = configuration.get('rust-client.channel', 'nightly');
+        this.channel = configuration.get('rust-client.channel', 'stable');
         this.componentName = configuration.get('rust-client.rls-name', 'rls');
 
         // Hidden options that are not exposed to the user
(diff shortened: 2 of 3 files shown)


Xanewok:
Thanks for the PR!

I agree that the current situation with rls not being available every nightly isn't ideal, but rather than setting it by default to stable, it'd be good to use user's default toolchain instead (as described in https://github.com/rust-lang-nursery/rls-vscode/issues/179).
There was work for this in https://github.com/rust-lang-nursery/rls-vscode/pull/180 but it's currently not continued anymore - is it something that you would be willing to work on?

kwonoj:
yeah, let me peek if I can make changes,

Xanewok:
@kwonoj ah my bad, I'm sorry. I went ahead and gave it a shot in the meantime at #220. However, if you're still willing to work on this or the nightly availability and rustup update issues in general, there is still work to do. :slightly_smiling_face: 

One notable improvement would be to detect if the rls-preview component is available and possibly detect what earliest nightly build has it. With this information we could warn the user that it's not available and suggest switching the channel to the earliest in which the rls is available.

Another good thing would be to start including [tests](https://code.visualstudio.com/docs/extensions/testing-extensions) for the extension itself, since it seems we stripped them some time ago. First possible good test would be to see if the `rustup show` [parsing](https://github.com/rust-lang-nursery/rls-vscode/pull/220/files#diff-a17328279d6a9110efb39a3a3e0410a4R180) in the #220 works as expected.

kwonoj:
closing as PR's now obsolete. I'll try another chance to contribute.

PR #220, opened 2017-12-30
Use local active toolchain as default for rust-client.channel

### LINKED ISSUE #179: Choose a channel for the RLS based on the current Rustup default.

Xanewok:
Fixes #179.

This allows the `'rust-client.channel'` option to be `null` and if so, the extension tries to infer the active channel from running `rustup show` in current workspace directory.

Manually tested with alternating stable/nightly toolchain and also with a toolchain override (servo repo), but it'd be good to also include a test for that.

r? @nrc

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -12,9 +12,8 @@
 
 import { workspace, WorkspaceConfiguration } from 'vscode';
 import { RevealOutputChannelOn } from 'vscode-languageclient';
-import * as child_process from 'child_process';
 
-import { parseActiveToolchain } from './rustup';
+import { getActiveChannel } from './rustup';
 
 function fromStringToRevealOutputChannelOn(value: string): RevealOutputChannelOn {
     switch (value && value.toLowerCase()) {
@@ -75,8 +74,9 @@ export class RLSConfiguration {
     }
 
     /**
-     * Tries to fetch the `rust-client.channel` configuration value. If it's null,
-     * then it tries to query active channel using rustup given at `rustupPath`.
+     * Tries to fetch the `rust-client.channel` configuration value. If missing,
+     * falls back on active toolchain specified by rustup (at `rustupPath`),
+     * finally defaulting to `nightly` if all fails.
      */
     private static readChannel(rustupPath: string, configuration: WorkspaceConfiguration): string {
         const channel = configuration.get<string | null>('rust-client.channel', null);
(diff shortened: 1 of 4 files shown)


src/rustup.ts
```diff
@@ -172,4 +172,27 @@ async function installRls(): Promise<void> {
...
 
     stopSpinner('RLS components installed successfully');
-}
\ No newline at end of file
+}
+
+/**
+ * Parses given output of `rustup show` and retrieves the local active toolchain.
+ */
+export function parseActiveToolchain(rustupOutput: string): string {
```
Xanewok: This isn't `async`, since it needs to finish to return a value with which global [`RLSConfiguration`](https://github.com/rust-lang-nursery/rls-vscode/pull/220/files#diff-8b01d5598b81332eb62385d26f3ead22R64) is created.
For the same reason it only parses the result and it's not a complete function that uses `CONFIGURATION.rustupPath` value, since it's called during `CONFIGURATION` construction.

As a side note, the configuration probably will need to be redesigned later on, as currently it's a global that's lazily constructed once and doesn't handle `'rust-client.*'` value changes

nrc: Could you add the 'side note' as a FIXME or file an issue please?

src/configuration.ts
```diff
@@ -58,15 +61,44 @@ export class RLSConfiguration {
...
     private static readRevealOutputChannelOn(configuration: WorkspaceConfiguration) {
         const setting = configuration.get<string>('rust-client.revealOutputChannelOn', 'never');
         return fromStringToRevealOutputChannelOn(setting);
     }
+
+    /**
+     * Tries to fetch the `rust-client.channel` configuration value. If it's null,
+     * then it tries to query active channel using rustup given at `rustupPath`.
+     */
+    private static readChannel(rustupPath: string, configuration: WorkspaceConfiguration): string {
```
nrc: I think this whole function should live in rustup.ts, rather than here.

Xanewok:
Done! Sorry it took so long, I forgot about it :disappointed: 

Can't think of a way to reliably test it (especially detecting local overrides) apart from using docker or a separate bash script that sets default toolchain to different one and tests the function output somehow with each change.

nrc:
Thanks!

PR #231, opened 2018-01-13
accept window/progress instead of beginBuild/diagnosticEnd

algesten:
This is the vscode side of https://github.com/rust-lang-nursery/rls/pull/653

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -129,7 +129,7 @@ function makeRlsProcess(): Promise<child_process.ChildProcess> {
     });
 
     return childProcessPromise.catch(() => {
-        window.setStatusBarMessage('RLS could not be started');
+        stopSpinner('RLS could not be started');
         return Promise.reject(undefined);
     });
 }
@@ -151,7 +151,7 @@ function startLanguageClient(context: ExtensionContext)
     }
     warnOnMissingCargoToml();
 
-    window.setStatusBarMessage('RLS: starting up');
+    startSpinner('RLS starting');
 
     warnOnRlsToml();
     // Check for deprecated env vars.
@@ -172,7 +172,7 @@ function startLanguageClient(context: ExtensionContext)
     // Create the language client and start the client.
     lc = new LanguageClient('Rust Language Server', serverOptions, clientOptions);
 
-    diagnosticCounter();
+    progressCounter();
 
     const disposable = lc.start();
     context.subscriptions.push(disposable);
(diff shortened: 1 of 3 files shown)


algesten:
screen shot 2018-01-14 at 17 28 43

Moved spinner order that it reads `RLS (spinner) <message>` so when `<message>` part changes`, the whole row doesn't flicker around.

src/extension.ts
```diff
@@ -205,17 +205,27 @@ async function autoUpdate() {
...
+        lc.onNotification(new NotificationType('window/progress'), function (progress: any) {
+            if (progress.done) {
+                delete runningProgress[progress.id];
+            } else {
+                runningProgress[progress.id] = true;
+            }
+            if (Object.keys(runningProgress).length) {
+                const status =
+                    typeof progress.percentage === 'number' ? asPercent(progress.percentage) :
+                    progress.message ? progress.message :
```
Xanewok: Title is ignored in favour of message, if it's specified - is that intended?
This leads to `RLS * [build]` followed by `RLS * crate_name` when building, where we lose information that we're building that crate.

algesten: @Xanewok This is indeed intended. I wanted to keep the information flow at the bottom terse, and it's mostly obviously what it's doing. The `[<title>]` is to get something rather than `RLS *`, but could maybe dropped. 

It's kinda interesting to see how long it spends doing `[diagnostics]` though. Already now I can see we should look into performance here.

algesten:
One thing I wondered is whether we should make the plugin in an interim handle both `beginBuild/diagnosticsEnd` and the new `progress` so that we can handle both old and new RLS?

Xanewok:
Since the user can specify the requested toolchain via the extension, I think there's no way around this other than just adding it in a backwards-compatibility fashion.
We could align releases with stable releases in general, but there's nothing stopping users from e.g. explicitly using stable Rust 1.23 and so it'd be best to keep the support for old messages.

algesten:
I fix

algesten:
@Xanewok @nrc I've rebased this off latest master. 

This incarnation handles both the old `beginBuild/diagnosticEnd` as well as the new `window/progress` to not have to be synced with rust nightly release of rls-preview.

src/extension.ts
```diff
@@ -207,17 +207,41 @@ async function autoUpdate() {
...
+
+        lc.onNotification(new NotificationType('window/progress'), function (progress: any) {
+            if (progress.done) {
+                delete runningProgress[progress.id];
+            } else {
+                runningProgress[progress.id] = true;
+            }
+            if (Object.keys(runningProgress).length) {
+                const status =
+                    typeof progress.percentage === 'number' ? asPercent(progress.percentage) :
```
nrc: Can you use an `if` statment rather than these nested ternary operators, it's pretty hard to read as is.

algesten: Ok.

algesten:
@nrc done

src/extension.ts
```diff
@@ -207,17 +207,45 @@ async function autoUpdate() {
     }
 }
 
-function diagnosticCounter() {
+function progressCounter() {
+    const runningProgress: { [index: string]: boolean } = {};
```
Xanewok: I'm really wonky with JS, but would it be possible to replace that with [`Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) for clarity?

Also not really sure why/how it's allowed to be `const`, since it's clearly modified in the callbacks, but I guess that's TypeScript shenanigans that I'm missing.

src/extension.ts
```diff
@@ -207,17 +207,45 @@ async function autoUpdate() {
...
+
+        stopSpinner('RLS');
+
+        lc.onNotification(new NotificationType('window/progress'), function (progress: any) {
+            if (progress.done) {
+                delete runningProgress[progress.id];
+            } else {
+                runningProgress[progress.id] = true;
+            }
+            if (Object.keys(runningProgress).length) {
```
Xanewok: If `Set` is used, these can be simplified (as per above)

src/extension.ts
```diff
@@ -207,17 +207,45 @@ async function autoUpdate() {
...
+                let status = '';
+                if (typeof progress.percentage === 'number') {
+                    status = asPercent(progress.percentage);
+                } else if (progress.message) {
+                    status = progress.message;
+                } else if (progress.title) {
+                    status = `[${progress.title.toLowerCase()}]`;
```
Xanewok: nit: I know it's up to clients to display this on the UI side, but shouldn't we just display it how servers send it?

algesten: Ultimately I don't really want to end up in this part of the logic. This is a last resort fallback. We should try to send percentages or messages instead. It looked really bad to me with the caps, but this just taste and style, so I'm only going to argue my case once :)

If you think it's an important point, I'll change it.

src/extension.ts
algesten: Agreed. `Set` is better.

`const` is not typescript shenanigans but javascript. It's not actually immutable (sadly), it's just the actual variable itself that is impossible to change. I.e.

```
const foo = 42;
foo = 43; // FAIL
```

algesten:
Let's land this :)

r? @Xanewok

algesten:
I just double checked that latest rls master works with the changes here. All good.

src/extension.ts
Xanewok: Right, makes sense. Thanks for the changes!

Xanewok:
Merged, thanks for working on this!

PR #234, opened 2018-01-14
Update the changelog to include 0.3.2 and unreleased changes.

allenbenz:
I think I got everything user visible that was not a documentation change.


nrc:
Awesome, thanks!

PR #235, opened 2018-01-14
Add snippets for all stable traits and macros in std and other common tasks

victorporof:
Signed-off-by: Victor Porof <victor.porof@gmail.com>

### DIFF
--- a/snippets/rust.json
+++ b/snippets/rust.json
@@ -306,6 +306,58 @@
         ],
         "description": "Add a test module"
     },
+    "fn": {
+        "prefix": "fn",
+        "body": [
+            "fn ${1:function}(${2:&self}) -> ${3:()} {",
+            "\t${4:unimplemented!()}",
+            "}"
+        ],
+        "description": "Add a function"
+    },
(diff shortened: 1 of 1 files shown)


snippets/rust.json
```diff
@@ -22,12 +245,12 @@
         ],
         "description": "Insert unreachable!"
     },
-    "println": {
```
victorporof: This macro wasn't removed, but grouped with others such as `write` and `format`.

nrc:
Hi, thanks for the PR and sorry to leave this PR without a comment for so long, it completely slipped off my radar. My personal preference has been to have a small, well-curated selection of snippets, rather than trying to be exhaustive. An informal twitter survey mostly supported that view. I think that the ideal solution is to have a separate extension with a large set of snippets which aims to be exhaustive for those users who like having lots of snippets, and keep the ones here a small set. We could cross-advertise the extensions to aid discovery. We could also host the snippet extension in the rust-lang org. If you're interested in that approach, let me know and we can make a plan.

Issue #237, opened 2018-01-21
RLS Cannot be started. 


neildanson:
As of 19/1/2018 (maybe earlier):

Extension says no rls installed, do I want to install. This is strange as rls is installed (via rustup) in cargo bin. 

Couldn't start client Rust Language Server
ErrorCould not install RLS component (rls)

VSCode:
Version 1.19.2
Commit 490ef761b76b3f3b3832eff7a588aac891e5fe80
Date 2018-01-10T15:55:03.538Z
Shell 1.7.9
Renderer 58.0.3029.110
Node 7.9.0
Architecture x64

Rust installed via rustup - 

toolchains installed:

stable-x86_64-pc-windows-gnu
stable-x86_64-pc-windows-msvc
beta-x86_64-pc-windows-msvc
nightly-x86_64-pc-windows-gnu
nightly-x86_64-pc-windows-msvc (default)

Uninstalled the plugin and reinstalling has no effect. Nothing about Rust or RLS in logs.

Not sure exactly what is wrong or how to diagnose more.

Xanewok:
What's your `rustc --version and Rust (rls) extension version?

neildanson:
rustc 1.25.0-nightly (15a1e2844 2018-01-20)

typing 

rls --version

results in 
error: toolchain 'nightly-x86_64-pc-windows-msvc' does not have the binary `rls.exe`

Even though I run it from the folder rls is installed in (.cargo/bin)

neildanson:
Vs Code plugin version Is 0.3.2

Xanewok:
You’re saying that RLS is installed via rustup - what does ‚rustup
component add rls-preview’ say?
The component changed its name some time ago, maybe you have the old
component installed?

neildanson:
It says 

info: component 'rls-preview' for target 'x86_64-pc-windows-msvc' is up to date

neildanson:
I uninstalled all gnu toolchains so that it was clearer what Rust (rls) would be using

Now I get 

"The Rust Language Server server crashed 5 times in the last 3 minutes. The server will not be restarted."

Xanewok:
Hm, that's weird. Would you mind setting a RUST_LOG env var to something like `rls=trace,rls_analysis=error` and then rerun VSCode? It should give more info about what RLS does and why it's crashing, the err output should be under Output > Rust Language Server window.

garretmh:
[...]

PR #240, opened 2018-01-26
Added check for most recent channel with RLS.

maccoda:
Wanted to have a shot at helping out on this one. I saw what the [Atom Rust IDE](https://github.com/mehcode/atom-ide-rust/tree/master/lib) has done for handling the RLS not being available on every nightly build so tried to take a leaf out of their book. 

This is my first time writing typescript so please let me know if I slipped up anywhere. 👍 

I am hoping this will assist on #237 and give the RLS some breathing room for getting into stable.

One thing that I didn't do but think is probably a useful next step is to make users aware that the configuration item for the channel being used is no longer what it used to start up and install the RLS. Let me know if you want me to have a go at that one, would be happy to help out.

### DIFF
--- a/src/rustup.ts
+++ b/src/rustup.ts
@@ -12,6 +12,7 @@
 
 import * as child_process from 'child_process';
 import { window, workspace } from 'vscode';
+import * as https from 'https';
 
 import { execChildProcess } from './utils/child_process';
 import { startSpinner, stopSpinner } from './spinner';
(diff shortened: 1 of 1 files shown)


nrc:
With the latest Rustup, we should never install a toolchain with a missing component without the users forcing the install, so I think we shouldn't need to do this. On the VSCode side, I think we should address the issue by  communicating what is going on to the user - i.e., if we `rustup update`, we only do so if there is an RLS, and if it is missing, we report that to the user (I believe the behaviour is currently correct, but we don't report a specific error to the user), if there is no RLS on the current channel, then we should recommend the user switch and tell them how to do that, but not actually do it, since they are probably in some weird dev scenario in this case (given the rustup change).

maccoda:
@nrc all sounds fair enough. Would a better approach then be to follow on from your last point there and provide a more user friendly notification of RLS not present over the current one which doesn't report it isn't found rather that it failed to install? Further, could perhaps add a section to the documentation to explain how to configure the channel.

Is that more in the direction of what you are looking for?

If so will rework this towards that 👍

nrc:
@maccoda, yes, that all sounds great, thanks!

maccoda:
Going to close this as I haven't had much time unfortunately and honestly really agreed with you that shouldn't touch too much of the user's set up and was realizing most of the things I was thinking of doing would become overly complex/messy. I will probably just post another PR hopefully soon just documenting what I had to do to get the plugin working when the RLS didn't ship as part of the nightly build 👍 Thanks for the help @nrc

PR #245, opened 2018-02-01
Support folding regions

lqd:
VSCode 1.17 added support for arbitrary markers to create foldable regions.
https://code.visualstudio.com/updates/v1_17#_folding-regions

This PR adds support for the `// region ...` `// endregion` comment markers to the language configuration.

### Before
image

### After
On the left: the regular state with the new "foldable region icons" in the gutter, and on the right: the same regions after being folded.

image


nrc:
Is this a common thing to do? It seems like adding editor-specific markers to generic source code and using semantically significant comments, both of which we would generally discourage. I'd want to see a strong precedent either in Rust or from other languages (languages which are not closely tied to a single editor, e.g., I've seen similar things in Swift, but that is only really expected to be used with XCode).

lqd:
A bit hard to tell, it seems common in vscode and generally MS languages where one could say Visual Studio was the only editor expected to be used. 
I've also seen it in Java with <editor-fold> in both Netbeans & Idea but I don't know if usage is common (and I don't remember if Eclipse supports it but I think it doesn't). 
No worrries!

PR #253, opened 2018-02-10
Handle promise correctly

kdy1:
- Don't drop promise.

Promise is similar to Future in rust, and dropping it results in a runtime error if promise fails. (It should be handled within 1 second).

`vscode` handles it correctly, and [allows extensions to return promise][vscode-promise]. 

 -  Use async for functions returning promise.

It removes one unexpected state because `async function`  cannot unwind.

[vscode-promise]:https://code.visualstudio.com/docs/extensionAPI/patterns-and-principles#_promises

### DIFF
--- a/src/utils/child_process.ts
+++ b/src/utils/child_process.ts
@@ -22,9 +22,11 @@ export interface ExecChildProcessResult<TOut = string> {
     readonly stderr: TOut;
 }
 
-export function execChildProcess(command: string): Promise<ExecChildProcessResult> {
+export async function execChildProcess(command: string): Promise<ExecChildProcessResult> {
     const r: Promise<ExecChildProcessResult> = new Promise((resolve, reject) => {
-        child_process.exec(command, (error, stdout, stderr) => {
+        child_process.exec(command, {
+            encoding: 'utf8',
+        }, (error, stdout, stderr) => {
             if (!!error) {
                 reject(error);
                 return;
@@ -37,4 +39,25 @@ export function execChildProcess(command: string): Promise<ExecChildProcessResul
         });
     });
     return r;
+}
+
+
+export async function execFile(command: string, args: string[], options: child_process.ExecFileOptions): Promise<ExecChildProcessResult> {
+    return new Promise<ExecChildProcessResult>((resolve, reject) => {
+        child_process.execFile(command, args, {
+            encoding: 'utf8',
+            ...options
+        }, (error, stdout, stderr) => {
+            if (!!error) {
+                reject(error);
+                return;
+            }
+
+            resolve({
+                stdout,
+                stderr,
+            });
+        });
+
+    });
 }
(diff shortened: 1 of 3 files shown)


nrc:
Hey, this is pretty hard to review since it includes quite a lot of reformatting as well as the meaningful changes. It would be great if you could separate this into two commits, or if that is difficult if you could point out where the meaningful changes are.

kdy1:
Sorry for a slow response. I was busy because of a national holiday.


Removed format changes and force pushed.

nrc:
Ping @saneyuki would you be able to review this PR please? It looks OK to me - aiui, mostly it is replacing explicit promises stuff with async/await, but I'm not confident with the JS here.

Thanks @kdy1 !

kdy1:
Fixed lints and rebased. r? @nrc

saneyuki:
@nrc I'm sorry that this reply is too late. I'll look this.

saneyuki:
> @nrc I'm sorry that this reply is too late. I'll look this.

I may need a few days to review (Sorry)

nrc:
@saneyuki thanks!

saneyuki:
@nrc I feel this changeset is nice.

--------------

@kdy1 Thank you!

kdy1:
@saneyuki Thank you for the review.

src/utils/child_process.ts
```diff
@@ -37,4 +39,25 @@ export function execChildProcess(command: string): Promise<ExecChildProcessResul
...
+        child_process.execFile(command, args, {
+            encoding: 'utf8',
+            ...options
+        }, (error, stdout, stderr) => {
+            if (!!error) {
+                reject(error);
+                return;
+            }
+
+            resolve({
```
nrc: This is giving me a type error when compiling:

```
[ts]
Argument of type '{ stdout: Buffer; stderr: Buffer; }' is not assignable to parameter of type 'ExecChildProcessResult<string> | PromiseLike<ExecChildProcessResult<string>> | undefined'.
  Type '{ stdout: Buffer; stderr: Buffer; }' is not assignable to type 'PromiseLike<ExecChildProcessResult<string>>'.
    Property 'then' is missing in type '{ stdout: Buffer; stderr: Buffer; }'.
```

kdy1: `encoding: 'utf8'` should make it's type string.

Did you set `encoding:??` in options argument? Typescript is dependently typed and tsc seems like doing inference across function boundaries.

Also making`encoding: 'utf8'` go last should fix the issue.

nrc: I'm just compiling a clean checkout, no changes or new functions. The options comes from `process.env`, I'm not sure how I would get `encoding` into that at compile time, perhaps it is platform-specific? The code should really compile across all platforms if that is the case.

kdy1: I guess typescript compiler is insisting that type cannot be determined (so defaulted to Buffer) because `process.env`  *may* have `encoding` field with other value. `tsc` is right in this case but I don't know why I didn't get any error. Moving `encoding: 'utf8'` to last will fix the issue.

kdy1: I'll send a pr soon.

Issue #92, opened 2017-07-20
Feature Suggestion: supply description, latest-version for Cargo.toml


chriskrycho:
This is *way* down the priority list, but I thought of it so I figured I'd put it here.

When you open a package.json file, you can hover over a dependency and it lists the short description and the latest version available for that package:

<img width="629" alt="hover" src="https://user-images.githubusercontent.com/2403023/28414687-d7d043e4-6d1a-11e7-8892-6b864c13ac57.png">

Adding this same functionality to this add-on for Cargo.toml files would be 💯 😍. Again: super low on the priority list, but a very nice-to-have feature.

nrc:
Agree this would be great to have!

kjeremy:
For reference I believe the equivalent functionality for package.json is implemented in https://github.com/Microsoft/vscode/blob/master/extensions/javascript/src/features/packageJSONContribution.ts

softprops:
Not sure if this exactly want you want but I've found https://marketplace.visualstudio.com/items?itemName=serayuzgur.crates to be useful in letting me know when I'm behind on crate versions and let's me know what the latest is

Xanewok:
I think we'd like to keep things lean here and rely on the external extension ([crates](https://marketplace.visualstudio.com/items?itemName=serayuzgur.crates) is great, I've been using it myself!) instead, so I'll close this issue.

PR #257, opened 2018-02-12
Display crates.io info in Hover for dependencies in Cargo.toml

clemensw:
Unpolished first attempt at #92:
* As it's my first time touching vscode, I don't know if I'm leaking Request Objects
* I assume crates.io would like some caching
* I don't know if you really want to add a dependency for this (and if so which http client library is acceptable)
* Horribly minimal heuristic to extract crate names: the possible semantics are extensive and I don't want to add real TOML parsing.  I'll see about adding at least a check for [dependencies].

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,6 @@
         "postinstall": "node ./node_modules/vscode/bin/install"
     },
     "dependencies": {
-        "request-light": "^0.2.2",
         "vscode-languageclient": "^3.4.2"
     },
     "devDependencies": {
@@ -177,6 +176,11 @@
                     "type": "string",
                     "default": "rls-preview",
                     "description": "Name of the RLS rustup component."
+                },
+                "rust-client.fetchCrateInfo": {
+                    "type": "boolean",
+                    "default": true,
+                    "description": "Fetch crate description from crates.io to display on Cargo.toml"
                 },
                 "rust.sysroot": {
                     "type": [
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
         "postinstall": "node ./node_modules/vscode/bin/install"
     },
     "dependencies": {
+        "request-light": "^0.2.2",
         "vscode-languageclient": "^3.4.2"
     },
     "devDependencies": {
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -36,6 +36,7 @@ export class RLSConfiguration {
     public readonly updateOnStartup: boolean;
     public readonly channel: string;
     public readonly componentName: string;
+    public readonly fetchCrateInfo: boolean;
     /**
      * Hidden option that can be specified via `"rls.path"` key (e.g. to `/usr/bin/rls`). If
      * specified, RLS will be spawned by executing a file at the given path.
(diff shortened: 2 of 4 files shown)


nrc:
Could you put this behind an option (default on) please? I know some people get twitchy about programs calling out to the internet.

I think we definitely need to cache the results, and I don't think that would be too much hassle.

I don't mind the dep, and don't have a preference, but I'm surprised there isn't a built-in facility - can we not use `fetch` in VSCode?

I'm not sure what to do about parsing TOML, I guess start with something very simple and we'll polish later if necessary.

clemensw:
Thanks so much for the feedback.

Options: I've got an option now.  Is rust-client the right namespace?

Caching: Caching results in memory now.  I believe they should stick for the life of the rls extension.  Do we need TTLs for long-running editors or serialized workspace state for caching longer than editor restarts?

Fetch Dependency: I've removed the dependency and lifted something based off the https api.

Toml parsing: I check that I'm in the [dependency] section now, but I'm not sure I like it.  I search for the section every time I hover over a crate candidate as I haven't figured out how to trigger a cache flush when the file is updated.  Also since I limit the hovers to that section I don't check for [dev-dependencies] or other plausible sections.  With the simpler heuristic we catch a few false positives (e.g. 'version  =' ).  I'm not sure which is worse.

nrc:
> Options: I've got an option now. Is rust-client the right namespace?

Yes

> Caching: Caching results in memory now. I believe they should stick for the life of the rls extension. Do we need TTLs for long-running editors or serialized workspace state for caching longer than editor restarts?

Sounds good, I don't think we need TTL - the data is quite small and I think people restart fairly often, in general

> Fetch Dependency: I've removed the dependency and lifted something based off the https api.

Sounds good.

> Toml parsing: I check that I'm in the [dependency] section now, but I'm not sure I like it. I search for the section every time I hover over a crate candidate as I haven't figured out how to trigger a cache flush when the file is updated. Also since I limit the hovers to that section I don't check for [dev-dependencies] or other plausible sections. With the simpler heuristic we catch a few false positives (e.g. 'version =' ). I'm not sure which is worse.

VSCode lets you watch files (we already watch Cargo.toml, but I think we register and watch from the server). You can nuke the whole cache rather than try anything clever.

Are there other plausible sections apart from dev-deps? Searching for the two sections seems fine. Alternatively, you could hard-code version as an exception to the simple approach. I don't mind which really, we don't need to be 100% here.

mgoszcz2:
This would be awesome to have working. Any plans on finishing and merging it? If not I'd love to make it work and submit a new PR if necessary.

nrc:
ping @clemensw are you able to get this across the line, or would you rather @mgoszcz2 or someone else take it over?

Xanewok:
I believe we'd like to focus on the RLS at its core here; also there is a popular [`crates`](https://marketplace.visualstudio.com/items?itemName=serayuzgur.crates) extension that focuses solely on that. I don't think it makes sense to compete with it atm and IIUC we can't offer more semantical knowledge here so I'd be inclined to close this PR now.

PR #271, opened 2018-03-12
Make encoding: 'utf8' last field.

kdy1:
So that the typescript compiler can use the fact that the value of `encoding` is always 'utf8'.

`vsce package` was successful.


nrc:
Thanks!

Issue #270, opened 2018-03-09
Compilation error in child_process.ts


nrc:
```
[ts]
Argument of type '{ stdout: Buffer; stderr: Buffer; }' is not assignable to parameter of type 'ExecChildProcessResult<string> | PromiseLike<ExecChildProcessResult<string>> | undefined'.
  Type '{ stdout: Buffer; stderr: Buffer; }' is not assignable to type 'PromiseLike<ExecChildProcessResult<string>>'.
    Property 'then' is missing in type '{ stdout: Buffer; stderr: Buffer; }'.
```

Introduced in #253

Xanewok:
It seems that the node typings are wrong?

For example the variants with args are: (notice the mismatch between function comments and stdout/stderr types)
```ts
    // `options` with `"buffer"` or `null` for `encoding` means stdout/stderr are definitely `Buffer`.
    export function execFile(file: string, options: { encoding: "buffer" | null } & ExecFileOptions, callback: (error: Error | null, stdout: string, stderr: string) => void): ChildProcess;
    export function execFile(file: string, args: string[] | undefined | null, options: { encoding: "buffer" | null } & ExecFileOptions, callback: (error: Error | null, stdout: string, stderr: string) => void): ChildProcess;

    // `options` with well known `encoding` means stdout/stderr are definitely `string`.
    export function execFile(file: string, options: { encoding: BufferEncoding } & ExecFileOptions, callback: (error: Error | null, stdout: Buffer, stderr: Buffer) => void): ChildProcess;
    export function execFile(file: string, args: string[] | undefined | null, options: { encoding: BufferEncoding } & ExecFileOptions, callback: (error: Error | null, stdout: Buffer, stderr: Buffer) => void): ChildProcess;
```

For args-less variant this is as expected, and we don't get an error there:

```ts
    // `options` with `"buffer"` or `null` for `encoding` means stdout/stderr are definitely `Buffer`.
    export function exec(command: string, options: { encoding: "buffer" | null } & ExecOptions, callback?: (error: Error | null, stdout: Buffer, stderr: Buffer) => void): ChildProcess;

    // `options` with well known `encoding` means stdout/stderr are definitely `string`.
    export function exec(command: string, options: { encoding: BufferEncoding } & ExecOptions, callback?: (error: Error | null, stdout: string, stderr: string) => void): ChildProcess;
```

[...]

PR #273, opened 2018-03-13
Actually use types for node.js 7.x

Xanewok:
Currently VSCode uses Node 7.9 version, so it's safer to pull types for this version instead. Thankfully the compile errors are gone here as well! Sorry for the churn and thanks @kdy1 for spotting the mistake!

r? @nrc

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     },
     "devDependencies": {
         "@types/mocha": "^2.2.43",
-        "@types/node": "^8.0.32",
+        "@types/node": "^7.0.56",
         "mocha": "^3.5.3",
         "tslint": "^5.7.0",
         "typescript": "^2.5.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -50,7 +50,7 @@
     },
     "devDependencies": {
         "@types/mocha": "^2.2.43",
-        "@types/node": "^8.0.32",
+        "@types/node": "^7.0.56",
```
Xanewok: The only versions of @types/node for node.js version 7.x are of `7.0.y` format, so this pulls the latest (which is marked in the source as targeting the major 7.x version)

kdy1:
Package version from my `node_modules/@types/node` was `8.xx`. That should be why I failed to reproduce #270 . After this pr, issue like #270 cannot occur again.

cc @nrc

nrc:
Thanks!

PR #275, opened 2018-03-16
Change `wait_to_build` default value to 1500ms

Xanewok:
RLS uses now 1500ms: https://github.com/rust-lang-nursery/rls/blob/242762ab99ac3df51b58bd83aa14fc9d83d0300a/src/main.rs#L71

See also https://github.com/rust-lang-nursery/rls/pull/771


nrc:
Thanks!

Issue #281, opened 2018-03-23
Add the `all_targets` option


nrc:
This is supported by the RLS but not by the VSCode plugin

Xanewok:
Fixed by #286.

PR #286, opened 2018-03-27
Add the rust.all_targets configuration value

mooman219:
Relevant issues #281 #261

The RLS added the `all_targets` option, which builds and indexes code for benches/tests/examples when the flag is set. Currently the RLS does none of that for code in those directories. This PR adds the configuration option in the rls-vscode plugin.


Xanewok:
Thank you!

PR #287, opened 2018-03-27
Update VSCode LSP client to 4.0

Xanewok:
Additionally fixes a bug where an abruptly closed server is left
hanging and is not restarted.

cc @sfackler

r? @nrc


nrc:
Thanks!

PR #302, opened 2018-04-25
Add more snippets

georich:
Crude(?) attempt at adding more snippets:

- I'm not sure exactly if/what I should enter for placeholders so I have been sparing
- I have arranged them in alphabetical order assuming there was no particular order before

I am also happy to add more, but am looking for feedback beforehand.

### DIFF
--- a/snippets/rust.json
+++ b/snippets/rust.json
@@ -9,32 +9,32 @@
         "description": "Insert for loop"
     },
     "fn": {
-	    "prefix": "fn",
-	    "body": [
-		    "fn ${1:name}($2) {",
-			"\t${0:unimplemented!()}",
-			"}"
-		],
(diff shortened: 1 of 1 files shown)


nrc:
Hey, sorry for leaving this so long. We've had similar PRs in the past and I honestly don't know what to do with them. When I've surveyed users, most say they want a smaller, more curated collection of snippets rather than a more complete collection.

I think it is better to close this PR, sorry. If you're interested, I think the right approach is to have extensions other than the RLS one which just provide snippets, e.g., https://marketplace.visualstudio.com/items?itemName=polypus74.trusty-rusty-snippets. So you might consider creating such an extension or contributing to that one.

Thanks for creating this PR and sorry to close it after silence for so long.

PR #306, opened 2018-04-30
[WIP] Redirect toolchain installation to an Output pane channel

### LINKED ISSUE #215: Put progress of operations into the VS Code Output window
Please consider creating an Output window channel for this extension where nightly toolchain installation stdout and RLS installation stdout would be redirected to. It went through fine for me, but I would like to see what's happening and I think overall it would increase transparency in a useful way.

TomasHubelbauer:
Closes #215. I haven't tested this yet, will do as time allows, but it should be close to done.

Anyone, please feel free to test this or improve upon it!

### DIFF
--- a/src/rustup.ts
+++ b/src/rustup.ts
@@ -85,7 +85,7 @@ async function tryToInstallToolchain(context: ExtensionContext): Promise<void> {
         // Collect error separately from the output channel as well to display an error message
         let error = '';
         await spawnChildProcess(
-            CONFIGURATION.rustupPath, ['toolchain', 'install', CONFIGURATION.channel],
+            CONFIGURATION.rustupPath, ['toolchain', '-v', 'install', CONFIGURATION.channel],
             stdOut => channel.append(String(stdOut)),
             stdErr => {
                 channel.append(String(stdErr));
--- a/src/rustup.ts
+++ b/src/rustup.ts
@@ -85,11 +85,11 @@ async function tryToInstallToolchain(context: ExtensionContext): Promise<void> {
         // Collect error separately from the output channel as well to display an error message
         let error = '';
         await spawnChildProcess(
-            CONFIGURATION.rustupPath, ['toolchain install ' + CONFIGURATION.channel],
-            stdOut => channel.append(stdOut),
+            CONFIGURATION.rustupPath, ['toolchain', 'install', CONFIGURATION.channel],
+            stdOut => channel.append(String(stdOut)),
             stdErr => {
-                channel.append(stdErr);
-                error += stdErr;
+                channel.append(String(stdErr));
+                error += String(stdErr);
             }
         );
 
--- a/src/rustup.ts
+++ b/src/rustup.ts
@@ -78,13 +78,14 @@ async function hasToolchain(): Promise<boolean> {
 async function tryToInstallToolchain(): Promise<void> {
     startSpinner('RLS', 'Installing toolchain…');
     try {
+        // TODO: Register this for disposal by passing the `context.subscriptions` array from `activate` all the way up here
         const channel = window.createOutputChannel('Rustup');
         channel.show();
-        // TODO: See if I need to split this into the path and the arguments
-        // TODO: Consider reporting error lines to the Problems pane?
         await spawnChildProcess(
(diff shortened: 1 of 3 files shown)


src/rustup.ts
```diff
@@ -78,9 +78,16 @@ async function hasToolchain(): Promise<boolean> {
 async function tryToInstallToolchain(): Promise<void> {
     startSpinner('RLS', 'Installing toolchain…');
     try {
-        const { stdout, stderr } = await execChildProcess(CONFIGURATION.rustupPath + ' toolchain install ' + CONFIGURATION.channel);
-        console.log(stdout);
-        console.log(stderr);
+        const channel = window.createOutputChannel('Rustup');
```
TomasHubelbauer: FYI this never gets disposed currently. I do not want to dispose of it here after the shell call is over because that would prevent the user from being able to inspect the channel after the toolchain installation. Instead, I think the channel should be returned to the called and added to the extension's `context.subscriptions` so it gets free'd with the extensions deactivation.

I _believe_ if the user was to initiate the toolchain installation twice (possible?), the channel will get reused if it already exists at `createOutputChannel` call-time. This needs to be verified. If yes, we shouldn't register it's destructor with `context.subsriptions` multiple times (so probably should instantiate the output channel outside of this method and pass it in - or it's `append` delegate only).

src/rustup.ts
```diff
@@ -78,9 +78,16 @@ async function hasToolchain(): Promise<boolean> {
...
+        channel.show();
+        // TODO: See if I need to split this into the path and the arguments
+        // TODO: Consider reporting error lines to the Problems pane?
```
nrc: I don't think we should do that since they are kind of meta-errors and don't have a line number in the project, etc. We could surface them as pop-up warnings/errors (like we do with overall success).

src/utils/child_process.ts
```diff
@@ -41,6 +41,15 @@ export async function execChildProcess(command: string): Promise<ExecChildProces
     return r;
 }
 
+export async function spawnChildProcess(command: string, onStdOut: (data: string) => void, onStdErr: (data: string) => void): Promise<number> {
+    const r: Promise<number> = new Promise((resolve, reject) => {
+        const process = child_process.spawn(command, []);
```
TomasHubelbauer: This may require splitting the command into the executable path and the command line arguments. See the comment in  [`rustup.ts` at line 85](https://github.com/TomasHubelbauer/rls-vscode/blob/0a8d07219b86ca921512561486760743d5b0d6e0/src/rustup.ts#L85).

src/rustup.ts
TomasHubelbauer: I've updated the code to instead collect the error separately in addition to outputting in in the Output channel and to display a window error notification message with the contents of the error, if any.

src/utils/child_process.ts
```diff
@@ -41,6 +41,16 @@ export async function execChildProcess(command: string): Promise<ExecChildProces
     return r;
 }
 
+export async function spawnChildProcess(command: string, args: string[], onStdOut: (chunk: Buffer | string) => void, onStdErr: (chunk: Buffer | string) => void): Promise<number> {
+    const r: Promise<number> = new Promise((resolve, reject) => {
+        const process = child_process.spawn(command, args);
+        process.stdout.on('data', onStdOut);
+        process.stderr.on('data', onStdErr);
+        // TODO: Figure out why this never resolves
```
TomasHubelbauer: This is the only outstanding problem now, the rest of the PR works as expected.

TomasHubelbauer:
As a part of working on this, I also considered it might be better to instead of spawning a child process to just run the command for the user in a new instance of the integrated terminal, but it doesn't look as though it is possible to listen for the invocation's result to be able to resolve the promise.

The reason this would be neat to have instead of the output channel is that the user could just rerun the command themselves and they would see all the fancy terminal stuff like progressbars etc. Right now while installing the toolchain, those are not visible, but it still gives a good indication of whether things are stuck or not.

TomasHubelbauer:
The mentioned VS Code API extension won't land as it was just marked out of scope so I will pick this up again without it.

PR #310, opened 2018-05-09
Change some options to match RLS

nrc:
Accompanies https://github.com/rust-lang-nursery/rls/pull/858

r? @Xanewok

### DIFF
--- a/package.json
+++ b/package.json
@@ -220,7 +220,7 @@
                         "null"
                     ],
                     "default": null,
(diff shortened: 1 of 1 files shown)


Xanewok:
Merging before https://github.com/rust-lang-nursery/rls/pull/858, since changing `all_targets` to true now is a good idea and I [think](https://github.com/rust-lang-nursery/rls/pull/858#pullrequestreview-118738756) it might be even possible to later remove the `build_*` if they're ignored.

PR #316, opened 2018-05-14
RLS's target_dir should be a subdirectory of target/

jonasbb:
Fixes https://github.com/rust-lang-nursery/rls/issues/864
This updates the fix in #312

### DIFF
--- a/package.json
+++ b/package.json
@@ -303,7 +303,7 @@
                         "string",
                         "null"
                     ],
-                    "default": "rls",
+                    "default": "target/rls",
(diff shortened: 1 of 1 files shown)


Xanewok:
Thank you! I should have been more careful when pushing the original fix.

Xanewok:
@nrc verified it works as expected, could you please publish a new minor version, so that hopefully less people might get affected by the issue?

nrc:
Published 0.4.4

PR #318, opened 2018-05-16
Add default channel setting & constrain channel values

### LINKED ISSUE #315: Add predetermined options for `rust-client.channel` config
Relevant config for this extension: https://github.com/rust-lang-nursery/rls-vscode/blob/309ad0cf5e6c3509b2fd08a4c59a6ba3987ac280/package.json#L175

An example of providing options: https://github.com/Microsoft/vscode/blob/401826a39951b9f7cffad6a571158710f5a027e5/extensions/css-language-features/package.json#L50-L54

LDSpits:
This pull requests adds autocompletion and a (normal) default for the ``rust-client.channel`` setting.

closes #315

### DIFF
--- a/package.json
+++ b/package.json
@@ -174,10 +174,14 @@
                 },
                 "rust-client.channel": {
                     "type": [
-                        "string",
-                        "null"
+                        "string"
                     ],
-                    "default": null,
+                    "enum": [
+                        "stable",
+                        "beta",
+                        "nightly"
+                    ],
+                    "default": "stable",
                     "description": "Rust channel to install RLS from."
                 },
                 "rust-client.rls-name": {
--- a/package.json
+++ b/package.json
@@ -174,15 +174,16 @@
                 },
                 "rust-client.channel": {
                     "type": [
-                        "string"
+                        "string",
+                        null
                     ],
                     "enum": [
                         "stable",
                         "beta",
                         "nightly"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -174,10 +174,14 @@
                 },
                 "rust-client.channel": {
                     "type": [
-                        "string",
-                        "null"
+                        "string"
                     ],
-                    "default": null,
```
nrc: I think changing the string to a set of options is a great thing to do, however, I think the default should still be `null` since in that case we'll use whichever channel the user would normally use and that is probably the expected behaviour (e.g., using nightly if the user would usually build with nightly).

LDSpits: Ahh, I didn't know that setting the value to ``null`` would do that! However, I do not think that ``null`` communicates "RLS updates itself based on the settings of your current toolchain config", can we maybe improve that somehow? 

I for instance thought that it would default to stable (and have thus overridden the channel for every project that I have)

nrc: We could add some documentation of that to the tool tip?

LDSpits: I've gone ahead and added some explanation to the description. If you want it somewhere else feel free to say so :)

nrc:
Thanks for the changes!

LDSpits:
No problem :)

PR #321, opened 2018-05-18
Add assert!() and assert_eq!() snippets.

OnlyLys:
As I understand it, RLS relies on Racer for snippet autocompletion [(rls/#827)](https://github.com/rust-lang-nursery/rls/issues/827). But since snippet autocompletion for macros hasn't been implemented yet [(rls/#829)](https://github.com/racer-rust/racer/issues/829), I propose we add two more commonly used macros: `assert!()` and `assert_eq!()` to rls-vscode as a temporary measure.

Later on, when macro autocompletion is implemented we should be able to remove these without breaking anything.

### DIFF
--- a/snippets/rust.json
+++ b/snippets/rust.json
@@ -36,6 +36,20 @@
         ],
         "description": "Insert println!"
     },
+    "assert": {
+		"prefix": "assert",
+		"body": [
+			"assert!($1)$0"
+		],
+		"description": "Insert assert!"
+	},
+	"assert_eq": {
+		"prefix": "assert_eq",
+		"body": [
+			"assert_eq!($1, $2)$0"
+		],
+		"description": "Insert assert_eq!"
+	},
     "macro_rules": {
         "prefix": "macro_rules",
         "body": [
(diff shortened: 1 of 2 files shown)


snippets/rust.json
```diff
@@ -36,6 +36,20 @@
         ],
         "description": "Insert println!"
     },
+    "assert": {
+		"prefix": "assert",
+		"body": [
+			"assert!($1)$0"
```
nrc: The syntax here and for assert_eq looks wrong, should it be `assert!($1, $0)`? Also I don't see why it is in this order and not `$0, $1`.

OnlyLys: Hi @nrc, [the `$0` tabstop in a snippet actually denotes the final cursor position](https://code.visualstudio.com/docs/editor/userdefinedsnippets#_snippet-syntax). Its purpose is to allow user to 'jump out' of the braces on the final 'tab' keypress. It's very convenient because it allows users to add a semicolon at the end without needing to manually move the cursor there.

But it's not only added there for convenience reasons, it's also to have parity with function snippet completions. If you're familiar with them, they all have a final tabstop immediately to the right of the closing brace.

nrc: Thanks for the explanation!

PR #329, opened 2018-05-31
1. Add tabstop for `println!` snippet. 2. Fix source code formatting for `assert!` and `assert_eq!` snippets.

OnlyLys:
### 1. Add tabstop at the end for `println!` snippet.
With reference to an earlier pull request [#321](https://github.com/rust-lang-nursery/rls-vscode/pull/321) which added `assert!` and `assert_eq!` snippets, I think the `println!` snippet should be slightly tweaked to have a tabstop at the end. 

### 2. Fix source code formatting for `assert!` and `assert_eq!` snippets.
My apologies for an earlier pull request [#321](https://github.com/rust-lang-nursery/rls-vscode/pull/321)  which used incorrect tab spacing in the source code. I have changed it to use 4-space tabs now.

### DIFF
--- a/snippets/rust.json
+++ b/snippets/rust.json
@@ -32,7 +32,7 @@
     "println": {
         "prefix": "println",
         "body": [
-            "println!(\"${1:{\\}}\", $0)"
+            "println!(\"${1:{\\}}\", $2)$0"
         ],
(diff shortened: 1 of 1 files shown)


nrc:
Thanks!

Issue #222, opened 2018-01-01
Incorrectly reports missing Cargo.toml if it's not in the root


chriskrycho:
I'm working on a small multi-lang project where the `Cargo.toml` is in a child of the root directory:

```
root/
  README.md
  rust/
    cargo.toml
  # other stuff
```

It incorrectly reports that:

> Cargo.toml must be in the workspace in order to support all features

Of course, `Cargo.toml` *is* in the workspace…

![code-error](https://user-images.githubusercontent.com/2403023/34464958-a61eb540-ee54-11e7-8356-d816fb6a244b.png)

jeffryang24:
Yes... It also happens to me...

maelvalais:
I think the error message has been changed to 

    A Cargo.toml file must be at the root of the workspace in order to support all features

meaning that rls-vscode expects cargo.toml to be at the root of your workspace. Sadly, this implies we have to find a workaround for multi-lang projects (for example, use the "Add folder to workspace" feature in vscode) 🤔

maelvalais:
I perused some of the extension's code but could not find how to hack it so that the 'root_path' for Cargo.toml is different from `workspace.rootPath`. That said, I found out that https://github.com/rust-lang-nursery/rls-vscode/commit/333332309695bb94e6d4e8c882db669d2b50d65f removed the variable `rls.root` that allowed running the rust language server from a different directory.

The commit message says: 
> rls.path is a better option, we don't need two ways and I don't think anyone uses this one any more

Maybe we need it back? 😊

enioluwa23:
Yup! Seems like that's it. Especially with neon/ffi-like modules I think people would like this customized instead of having two VS Code workspaces open at the same time to have rls functionality.

joseluisq:
Any progress on this feature?

jannickj:
This issue is really annoying when working with a rust, rarely is rust my sole project. 
@maelvalais Solutions seems like a nice hack, but it doesn't fix the problem if you have multiple rust projects at the same time. What we actually need is for the extension to find the cargo.toml based on the file you have active and then either reuse or spin up a new language server.

maelvalais:
You are right, my hack does not really help in these scenarios. Let us imagine we have open vscode in an hypothetical `project/` folder that contains multiple rust projects (each one with a `Cargo.toml` and `src/*.rs`):

```
project/
├── lib1/
│  ├── Cargo.toml
│  └── src/
│     └── main1.rs
├── lib2/
[...]

PR #334, opened 2018-06-07
Allow having Cargo.toml in a subdirectory with `rust-client.cargoSubdir`

maelvalais:
Fixes https://github.com/rust-lang-nursery/rls-vscode/issues/222.

This patch is quite hacky and may reveal my lack of knowledge on the RLS or
LSP architectures. My idea was to give the user the opportunity using the
variable `rust-client.cargoSubdir` to select a subdirectory as the root for
the Cargo.toml file.

To do that, I hacked the vscode's workspaceFoldervariable; my hack won't
work with multi-workspaces.

Alternate solutions:
- (not so good idea) add a new setting to the initialize JSON-RPC (similar
  to the existing `omitInitBuild` in the rls server) that would set a
  different path that the one given by vscode's workspaceFolder.
- (better) we add support for multi-root workspaces in the RLS server;
  (mentionned in  <https://github.com/rust-lang-nursery/rls/issues/608>)
  but this does not really help when we want to open a general project
  where one of the subfolders is a rust project.
- we tell people that in order to get RLS working in a subdirectory, they
  must use the 'Multi-crate projects' feature or the 'rlsCommandOverride'
  feature (see README at <https://github.com/mehcode/atom-ide-rust>).

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,8 @@
         "postinstall": "node ./node_modules/vscode/bin/install"
     },
     "dependencies": {
-        "vscode-languageclient": "^4.1.4"
+        "vscode-languageclient": "^4.1.4",
+        "path": "^0.12.0"
     },
     "devDependencies": {
         "@types/mocha": "^2.2.43",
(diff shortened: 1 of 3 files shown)


nrc:
I recently landed support for multi-root workspaces in the VSCode extension (using one language server per workspace folder, rather than just one). Does that let you add a per-folder setting to do this?

I think that just having one setting for all projects is a bit of a blunt instrument for this.

Xanewok:
Support has landed for the multi-root workspaces and it seems to work quite well now, so I'd like to close this PR if you don't mind.

Issue #247, opened 2018-02-02
execChildProcess does not work with wsl


kdy1:
https://github.com/rust-lang-nursery/rls-vscode/blob/5afe3bbe5b6cf538cafea9bf8bfe3b3779026041/src/utils/child_process.ts#L25-L40

For native windows binaries, this works well. But when wsl is involved (even if it's subprocess invoked from simple windows exe file) returned promise does not resolve because callback passed to `child_process.exec` isn't called. I'm not sure if this is an issue of rls-vscode, but this can be workarounded.

Also, I'm sure that command itself is invoked, because `stdout` is set to `nightly-x86_64-unknown-linux-gnu (default)` if I set timeout. (but timeout error occurs)

-----

When I modify definition like below and invocations, it works. 

```typescript

export async function execChildProcess(command: string, args: string[]): Promise<ExecChildProcessResult> {
    // FIXME: async
    const child = child_process.spawnSync(
        command, args, { encoding: 'utf8' },
    );

    if (child.error) {
        throw new Error(`Error running `${command}`: ${child.error}`);
    }

    if (child.status > 0) {
        throw new Error(`Error getting sysroot from `${command}`: exited with \`${child.status}\``);
    }

    if (!child.stdout || typeof child.stdout.toString !== 'function') {
        throw new Error(`Couldn't get sysroot from `${command}`: Got no ouput`);
    }

    return {
        stdout: child.stdout,
        stderr: child.stderr,
    };
}
```

So.. child_process.exec` seems like a root of issue. This should be replaced with `child_process.spawn` to work with wsl.

yisonPylkita:
Yep, you are right. GH-335 will fix this issue

yisonPylkita:
https://github.com/rust-lang/rls-vscode/pull/488 was merged a few days ago. Thus this bug should be fixed.

Do you mind running this extension from `master` to confirm the fix ?

kdy1:
@yisonPylkita I uninstalled wsl for some reasons. So I can't.

Xanewok:
This was done as part of the 0.6 refactor, closing this now.

Issue #144, opened 2017-09-08
Possible to use this ext where rustup/rustc/cargo is provided by WSL?


kardeiz:
Is it possible to use this extension on Windows, where VSCode is a Windows app, but where `rustup`/`rustc`/`cargo`/`rls` are provided through the Windows Subsystem for Linux (WSL) `bash`?

I've tried some naive things, like changing the integrated terminal and `windowsExec` to use `bash`, which works just fine, but the plugin keeps complaining "Couldn't start client Rust Language Server".

I also tried hardcoding `rls.path` to the path of RLS (installed via rustup on WSL bash).

Thanks!

Xanewok:
Hm, piping stdio/stdout `bash -c '<cmd>'` does seem to work as per [Windows Interoperability](https://msdn.microsoft.com/en-us/commandline/wsl/interop).

I believe https://github.com/Microsoft/vscode/issues/22663 might be relevant (as it gets into internals how language server processes are spawned)? I believe the issues it that node runtime in VSCode spawns processes naively using the same subsystem, which might be causing problems. It seems someone posted a workaround, could you try and see if it works?

kardeiz:
Sorry, I just gave up on this for a while. If you are referring to the `wsl-bash-tools` wrapper mentioned in that thread, I tried that, but it doesn't work (and possibly doesn't quite address the issue of child process forking? This is mostly over my head).

I just discovered that it's possible to view the Chrome dev console in VSCode, which was a huge help. 

I see in the warnings:

```
Error: Error running `rustc`: Error: spawnSync C:\Users\me\vscode_ext\wsl-bash-tools-master\wsl-bash-run.bat /home/me/.cargo/bin/rustup ENOENT
	at getSysroot
```

Then I see:

```
running C:\Users\me\vscode_ext\wsl-bash-tools-master\wsl-bash-run.bat /home/me/.cargo/bin/rls
```

And no further errors or output, although the extension host crashes 30 seconds or so later.

Any thoughts?

jonasbb:
Just an idea to establish connection between RLS and VS Code. I cannot test this as I don't use Windows. Can you run

    nc -kl `which rls` 1234

in WSL. Under Windows you would then create a cmd file containing the equivalent of `nc localhost 1234` and configure VS Code by setting the `rls.path` configuration to this cmd script.

[...]

PR #335, opened 2018-06-12
WSL support

yisonPylkita:
To resolve all concerns with WSL (#160, #247, #144 ) and to fix my half-baked implementation GH-161 I have created this PR.  It should fix all of the problems with running this extension via WSL.

This is still WIP. The only missing thing is mapping filesystem paths from vscode language client (windows) to rls (wsl)

Example: 
I have simple rust application `kudu`. Lets run vscode on this application project.
Initialization, rls detection, (installation if needed) and so on works without problems. Hard part begins when rls tries to look for `Cargo.toml`. 

I got error 
```
Error reading manifest path: ErrorMessage { msg: "could not find `Cargo.toml` in `/c:/H_DEV/kudu` or any parent directory" }
```

It seems that to make this whole WSL thing working we need to make rls aware of wsl. Simplest way to fix it should be by introducing a filesystem paths mapping mechanism. So path `c:/H_DEV/kudu` should be mapped to `/mnt/c/H_DEV/kudu`. 

@Xanewok any ideas how to implement paths mapping in rls ?

### DIFF
--- a/package.json
+++ b/package.json
@@ -148,6 +148,11 @@
                     "default": "rustup",
                     "description": "Path to rustup executable"
                 },
+                "rust-client.wslPath": {
+                    "type": "string",
+                    "default": "",
+                    "description": "Path to WSL executable"
+                },
                 "rust-client.rlsPath": {
                     "type": [
                         "string",
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -31,6 +31,7 @@ function fromStringToRevealOutputChannelOn(value: string): RevealOutputChannelOn
 
 export class RLSConfiguration {
     public readonly rustupPath: string;
+    public readonly wslPath: string | undefined;
     public readonly logToFile: boolean;
     public readonly revealOutputChannelOn: RevealOutputChannelOn = RevealOutputChannelOn.Never;
     public readonly updateOnStartup: boolean;
@@ -49,6 +50,7 @@ export class RLSConfiguration {
 
     private constructor(configuration: WorkspaceConfiguration) {
         this.rustupPath = configuration.get('rust-client.rustupPath', 'rustup');
+        this.wslPath = configuration.get('rust-client.wslPath', undefined);
         this.logToFile = configuration.get<boolean>('rust-client.logToFile', false);
         this.revealOutputChannelOn = RLSConfiguration.readRevealOutputChannelOn(configuration);
         this.updateOnStartup = configuration.get<boolean>('rust-client.updateOnStartup', true);
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -22,7 +22,7 @@ import { commands, ExtensionContext, IndentAction, languages, TextEditor,
     TextEditorEdit, window, workspace } from 'vscode';
 import { LanguageClient, LanguageClientOptions, Location, NotificationType,
     ServerOptions } from 'vscode-languageclient';
-import { execFile, ExecChildProcessResult } from './utils/child_process';
+import { ExecChildProcessResult, run_rustup } from './utils/child_process';
 
 // FIXME(#233): Don't only rely on lazily initializing it once on startup,
 // handle possible `rust-client.*` value changes while extension is running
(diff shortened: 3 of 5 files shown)


nrc:
> It seems that to make this whole WSL thing working we need to make rls aware of wsl. Simplest way to fix it should be by introducing a filesystem paths mapping mechanism. So path c:/H_DEV/kudu should be mapped to /mnt/c/H_DEV/kudu

Where do these paths come from? Are they the ones from the compiler? I.e., the filenames in spans? Or is it something more specific to the RLS?

yisonPylkita:
These paths comes from `vs-code` but an extension has no way to modify them. 
I will clarify myself - they comes from `vs-code` language server. 

I suspect that first problematic path comes from initialize request - https://microsoft.github.io/language-server-protocol/specification#initialize

nrc:
So the file URIs in the ranges from the client?

bharrisau:
Inside the WSL environment there is the `/bin/wslpath` helper that will map from Windows paths to Linux paths. But there doesn't seem to be a good place to hook that path translation in.

rdrgn:
So using `wslpath` like this?
```
C:\>wsl.exe wslpath 'c:/H_DEV/kudu'
/mnt/c/H_DEV/kudu

C:\>wsl
remin@laptop:/mnt/c$ wsl.exe wslpath 'c:/H_DEV/kudu'
/mnt/c/H_DEV/kudu
```

Of course, we do not have to call `wsl.exe` in WSL.

```
remin@laptop:/mnt/c$ wslpath 'c:/H_DEV/kudu'
/mnt/c/H_DEV/kudu
```

bharrisau:
Yes, it should also work in reverse with the `-w` flag (from memory).
Better than hard coding a regexp or something as wslpath looks at the wsl
configuration to do the mapping correctly (in case you use a non standard
mount point).

tim-stasse:
So what needs to be done to get this across the line? I'd love to have this working, but this PR has been open since June...

bharrisau:
It seems like the communication is between VSCode and RLS(in WSL) directly. I can't see anywhere that the rls-vcode extension is able to map the paths before they are sent to RLS. That is the main part missing from this PR. If someone can point me in the right direction I can take a look. All the other language extensions that handle WSL (that I've looked into anyway) are using a custom protocol and aren't using the LanguageServerClient stuff.

This PR currently just allows `rustup` et.al. to be used inside WSL - I'm doing the same thing by just using a `rustup.bat` file and setting the `rustupPath` config to point to it. Same thing.

yisonPylkita:
> All the other language extensions that handle WSL (that I've looked into anyway) are using a custom protocol and aren't using the LanguageServerClient stuff

This is interesting. I didn't had enough time to look into different vs-code plugins with WSL support.
Yet I think that replacing `vs-code` protocol just for the sake of this functionality is too much.

Originally I came up with an idea to run RLS (inside WSL) with an command line option like `--wsl-mapping` and to make RLS translate paths from Windows to WSL paths (So `C:\\Users\user1\rust_project` would become `/mnt/c/Users/user1/rust_project`)

My rust skills are too low to do this so I gave up on this

dtysky:
Any update?

yisonPylkita:
@dtysky Pretty much done. Functionality is implemented yet it still require some tests.

Do you mind to be a tester ? (Windows 10 required)

dtysky:
> @dtysky Pretty much done. Functionality is implemented yet it still require some tests.
> 
> Do you mind to be a tester ? (Windows 10 required)

My pleasure, so what should I do?

Xanewok:
@dtysky To run the extension from source (and this PR branch) please follow the steps in [contributing.md](https://github.com/rust-lang/rls-vscode/blob/master/contributing.md).

I don't develop on Windows, but I believe the use case here is that you want to install and use your Rust toolchain from inside the Windows Subsystem for Linux (WSL) inside a Visual Studio Code instance that's installed in and managed by Windows.

To verify the fix, I believe it's enough to simply create a dummy project with `cargo new my-lib`, initialize VSCode inside and check if you get any completion or IDE features whatsoever.

@yisonPylkita is the above correct and/or did I miss anything?

Also, I'm going to close this PR since the work on this moved to #488, please move the discussion there.

yisonPylkita:
@Xanewok You are right. These are the basic steps to check this feature.

I wrote a little tutorial how to test this extension https://github.com/rust-lang/rls-vscode/pull/488#issuecomment-461301456. Hope it will be helpful for anybody who wants to test this feature.

PR #348, opened 2018-06-24
Streamline dev workflow

matklad:
I am not entirely sure what I am doing here, but I think this should allow push-button installation of the extension for developers, while avoiding `npm install -g` which might not be a great idea because it modifies global environment.


nrc:
lgtm, thanks!

PR #351, opened 2018-06-26
Recomend tslint extension for this workspace

matklad:
Currently, it's very easy to make the linter unhappy (just forget a
semicolon), and then it's very hard to figure what's wrong (linter
errors are not shown in the editor, linter is only run by
`installDevExtensions`, and there its error messages are swallowed)

This commit adds tslint as a recommended extension for this VS Code
workspace. That is, new contributors will be prompted to install tslint
extension, and will see errors inline afterwards.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,5 @@
+{
+	"recommendations": [
+		"eg2.tslint"
+	]
+}


.vscode/extensions.json
```diff
@@ -0,0 +1,7 @@
+{
+	// See http://go.microsoft.com/fwlink/?LinkId=827846
+	// for the documentation about the extensions.json format
```
nrc: Comments are not usually permitted in JSON files, even if VSCode is OK with them, it's probably best to leave it out.

matklad: Fixed!

nrc:
Thanks!

PR #352, opened 2018-06-26
Don't overwrite LD_LIBRARY_PATH

matklad:
Some users need their LD_LIBRARY_PATH to keep their
system from falling appart.


Xanewok:
Indeed, seems like the right thing to do (and what should've been done in the first place). Thanks!

PR #360, opened 2018-07-06
Add setting for clippy_preference

### LINKED ISSUE #298: Add the `clippy-preference` option
This seems to be supported by RLS as of [this commit](https://github.com/rust-lang-nursery/rls/commit/f7031d94562e64135d0dbc90b24004d79145ab9e), but not by the VSC plugin.

jechase:
Resolves #298.

By inspecting the vscode log files, I can confirm that it's receiving the setting properly, but I've yet to see any actual clippy lints when I have it set to "on."

Is there something else that needs to be done or am I just on a bad rls version?

### DIFF
--- a/package.json
+++ b/package.json
@@ -311,6 +311,17 @@
                     "description": "Enables code completion using racer.",
                     "scope": "resource"
                 },
+                "rust.clippy_preference": {
+                    "type": "string",
+                    "enum": [
+                        "off",
+                        "opt-in",
+                        "on"
+                    ],
+                    "default": "opt-in",
+                    "description": "Controls eagerness of clippy diagnostics when available.",
+                    "scope": "resource"
+                },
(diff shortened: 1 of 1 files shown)


nrc:
> Is there something else that needs to be done or am I just on a bad rls version?

Clippy is an optional feature and we don't build it by default. To get Clippy you need to build the RLS yourself and enable the Clippy feature at compile time. This should change in the near future though.

nrc:
So unfortunately this PR won't do anything until we do that. CLosing for now.

jechase:
Ah, ok, well that explains my trouble getting it to work. From [this thread](https://github.com/rust-lang-nursery/rls/issues/149), it sounded like if it compiled, clippy would be included in RLS.

nrc:
Clippy should be included from the next nightly

PR #365, opened 2018-07-11
Fix active toolchain detection

### LINKED ISSUE #305: Why does it install Nightly?
I installed this vscode extension. It is nice! However, I noticed: When I first installed it, it installed first rust nightly, then it installed RLS. It was as if the vscode setup believed that nightly was necessary in order to boot RLS. However, a Rust user I talked to informed me that RLS has not required nightly since Rust 1.21. The RLS instructions https://github.com/rust-lang-nursery/rls suggest only that you need to install the most current version.

Is the Rust Nightly install actually necessary, or is this just a vestigial quirk of when RLS required nightly?

segevfiner:
This fixes the parsing of `rustup show`, and also uses the new `rustup show active-toolchain` (https://github.com/rust-lang-nursery/rustup.rs/pull/1449) when available, which is meant for use from scripts/code.

Fixes #305

### DIFF
--- a/src/rustup.ts
+++ b/src/rustup.ts
@@ -225,9 +225,17 @@ export function getActiveChannel(rustupPath: string, wsPath: string): string {
     // rustup info might differ depending on where it's executed
     // (e.g. when a toolchain is locally overriden), so executing it
     // under our current workspace root should give us close enough result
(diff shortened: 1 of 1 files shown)


Xanewok:
Great to see that `rustup show active-toolchain` is now a thing! Previous way was pretty hacked up... Thanks!

PR #366, opened 2018-07-12
Goto Definition now works with files outside of the workspace

aloucks:


aloucks:
You're right, it does run into issues with multiple workspace folders. I wonder if you could run another copy of RLS and the language client that's using a documentSelector on `$HOME/.cargo/registry/src/**/*`?

nrc:
>  I wonder if you could run another copy of RLS and the language client ...

That's a great idea. I have no idea whether it would work or not tbh. I imagine it might have a problem since we don't want to build std, we want to use already existing save-analysis data. Doing that properly would require some modification to the RLS. Would be a fun little project though (and useful for other things, actually).

jimliang:
We neet it! @nrc

PR #367, opened 2018-07-12
Upgrade vscode-languageclient to 4.3.0

aloucks:
This fixes issues where tooltip sections aren't rendered correctly.

The previous behavior omitted the thin black lines between sections and caused content to run together unless additional newlines were introduced.


Xanewok:
I assume this (https://github.com/Microsoft/vscode-languageserver-node/commit/14bc143939768799c7b7873553dd22b9a12f7823) is the relevant change? Thanks!

PR #369, opened 2018-07-14
Use standard LSP 3.6 Goto implementation command

### LINKED ISSUE #368: Goto implementation command needs updating
As of [this](https://github.com/rust-lang-nursery/rls/pull/936) in the rls repo the rls now uses the standard `textDocument/implementation` command instead of `rustDocument/implementations`. The menu item provided by this extension should be fixed to use the correct method (either via configuration or some other manner)

Xanewok:
Fixes #368.


dgriffen:
Should this be configurable for users on older versions of the rls or is that not a big issue?

Xanewok:
You're right, I think it'd be nice to use a fallback during the transition period; I'll add a commit with it

Xanewok:
The change is small and I tested with old and new RLS versions, so I'm gonna go ahead and merge this.
@dgriffen thanks again for making the RLS use the officially supported now message!

PR #370, opened 2018-07-15
Added print! macros snippet and some fixes

enzovitaliy:


snippets/rust.json
```diff
@@ -29,10 +29,17 @@
...
+        "body": [
+            "print!($1)$0"
+        ],
+        "description": "Insert print!"
+    },
     "println": {
         "prefix": "println",
         "body": [
-            "println!(\"${1:{\\}}\", $2)$0"
+            "println!($1)$0"
```
nrc: I don't think this is a great fix since I think it is nice to have the `"", ` there for the user and have the tab stops for inside and outside. However, you're probably right that having the `{}` is not necessary.

PR #382, opened 2018-07-23
Changes the readonly fields in the

### LINKED ISSUE #233: Handle rust-client.* configuration value changes
Currently we lazily initialize the configuration once, at extension startup and later we don't update it, even on related `rust-client.*` configuration value updates.
Probably lazily initializing it from initial user settings is good, but we need to also update changing the values on the go (e.g. changing toolchain used), see https://github.com/rust-lang-nursery/rls-vscode/pull/220#discussion_r159567580.

chuck-flowers:
Fix #233 Changes the readonly fields in the RLSConfiguration 
to get-only properties that calculate the values lazily 
when the properties are read.

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -30,16 +30,51 @@ function fromStringToRevealOutputChannelOn(value: string): RevealOutputChannelOn
 }
 
 export class RLSConfiguration {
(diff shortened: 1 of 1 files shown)


nrc:
Thank you!

PR #423, opened 2018-09-06
Improve documentation of the clippy_preference setting.

### LINKED ISSUE #417: The meaning of the "opt-in" setting for rust.clippy_preference is unclear.
The `rust.clippy_preference` setting is defined in `package.json` as

    "rust.clippy_preference": {
        "type": "string",
        "enum": [
            "on",
            "opt-in",
            "off"
        ],
        "default": "opt-in",
        "description": "Controls eagerness of clippy diagnostics.",
        "scope": "resource"
    },

This gives no hint how to opt in if that option is selected.  I suggest copying the description from the RLS documentation instead:

> Controls eagerness of clippy diagnostics when available. Valid values are _(case-insensitive)_:
>   - `"off"` Disable clippy lints.
>   - `"opt-in"` Clippy lints are shown when crates specify `#![warn(clippy)]`.
>   - `"on"` Clippy lints enabled for all crates in workspace.

smarnach:
Fixes #417.

There doesn't seem to be any way to include markup in the description.  I used back ticks to delimit code anyway, since I think it improves clarity, and it is done at other places in the same file.

The new description renders like this for me on VS Code 1.27.1:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -341,7 +341,7 @@
                         "off"
                     ],
                     "default": "opt-in",
-                    "description": "Controls eagerness of clippy diagnostics.",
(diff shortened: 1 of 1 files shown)


nrc:
Thank you! And sorry it took a while to merge.

PR #431, opened 2018-09-26
Adds an option for disabling Rustup

bkchr:
If Rustup is disabled, rls/rustc will be used from PATH.
This is quite useful for systems that do not use any Rustup.


nrc:
Thanks!

PR #457, opened 2018-11-28
search for the cargo toml based on the current file

jannickj:
Code for the proposed fix: #222

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -56,6 +56,7 @@ function didOpenTextDocument(document: TextDocument, context: ExtensionContext):
     folder = getOuterMostWorkspaceFolder(folder);
     folder = getCargoTomlWorkspace(folder, document.uri.fsPath);
     if (!folder) {
+        window.showWarningMessage(`A Cargo.toml was not found for document ${ document.uri.fsPath }`);
         return;
     }
 
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -56,7 +56,7 @@ function didOpenTextDocument(document: TextDocument, context: ExtensionContext):
     folder = getOuterMostWorkspaceFolder(folder);
     folder = getCargoTomlWorkspace(folder, document.uri.fsPath);
     if (!folder) {
-        window.showWarningMessage(`A Cargo.toml was not found for document ${ document.uri.fsPath }`);
+        stopSpinner(`RLS: Cargo.toml missing`);
         return;
     }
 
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -53,6 +54,10 @@ function didOpenTextDocument(document: TextDocument, context: ExtensionContext):
         return;
     }
     folder = getOuterMostWorkspaceFolder(folder);
+    folder = getCargoTomlWorkspace(folder, document.uri.fsPath);
+    if (!folder) {
+        return;
```
nrc: Can/should we issue a warning about missing Cargo.toml here?

jannickj: yes probably a good idea

jannickj: My only concern is that you will get a pop up if just opening a .rs file outside it's project

jannickj:
Changed how the message is shown from a pop up to a nice little message in the bottom, much more pleasant from a UX perspective.

src/extension.ts
jannickj: Okay I've come up with a solution that will work in both use cases.

maelvalais:
Thank you, @jannickj!!!  🎉 🍾

rickvanprim:
@nrc @jannickj I see this changed was backed out.  Any context on why?  Will this feature be re-added soon?

Xanewok:
IIRC this introduced a bug that broke project detection and it did impact a lot of users.

Could you remind me what this aims to fix? Not warning against missing Cargo.toml when the user opens a project that has a Rust project nested and not at the top level?

I'll try to tackle this again in the immediate future.

rickvanprim:
I can't speak for others, but for me this being able to have a project that contains one or more Rust crates mixed in with other projects, where Cargo is not the root build system and Rust is not the only language.  RLS detecting Cargo.toml relative to the source file being looked at means the crate can be located anywhere in the repo, not just at the root.

rickvanprim:
@Xanewok any updates?

jannickj:
@Xanewok The problem is with having multiple rust project in the space workspace.

The current setup is that the `Cargo.toml` must be in the root of the workspace. However with this change it would allow the project to work in any subfolder.

@rickvanprim I remember the problem being impact on windows users. I use linux exclusively and hadn't properly tested nor forwarned that it probably should be tested.

I will see if I can fix the compatibility problems, and then try again and hopefully we can have others help the testing.

PR #459, opened 2018-11-29
Run RLS with cwd set to crate root (fixes #448)

### LINKED ISSUE #448: RLS runs with the home directory as the CWD, breaking RLS wrappers
When RLS is started by this plugin, it's run with the home directory as the current working directory, as opposed to the crate root. This breaks wrapper scripts that rely on project-specific information to run RLS in a particular environment or with a particular configuration.

For example, this is what I use for running RLS in the context of a Nix shell (essentially a self-contained environment with all the project dependencies present), under [`ide-rust`](https://github.com/rust-lang-nursery/atom-ide-rust) in Atom:

```sh
if [ -e shell.nix ]
then
    nix-shell --run "rls $@"
else
    rls $@
fi
```

That script is then configured as the RLS path, and that will result in RLS being run within the correct environment. The same script does not work with this plugin, since the current working directory is always `/home/sven` for me no matter the project, and therefore the build gets executed in the wrong environment (and subsequently fails).

Changing this to always pick the crate root as the cwd would resolve this issue, and I don't think it would break anything else. I'd also argue that it'd make more sense semantically, since that's how one would run it from a shell.

joepie91:
Title is self-explanatory :)

Edit: For linkability of the corresponding issue, #448.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -371,16 +371,17 @@ class ClientWorkspace {
     async makeRlsProcess(): Promise<child_process.ChildProcess> {
         // Allow to override how RLS is started up.
         const rls_path = this.config.rlsPath;
+        const cwd = this.folder.uri.fsPath;
 
         let childProcessPromise: Promise<child_process.ChildProcess>;
         if (rls_path) {
             const env = await this.makeRlsEnv(true);
             console.info('running ' + rls_path);
-            childProcessPromise = Promise.resolve(child_process.spawn(rls_path, [], { env }));
+            childProcessPromise = Promise.resolve(child_process.spawn(rls_path, [], { env, cwd }));
         } else if (this.config.rustupDisabled) {
             const env = await this.makeRlsEnv(true);
             console.info('running rls from $PATH');
-            childProcessPromise = Promise.resolve(child_process.spawn('rls', [], { env }));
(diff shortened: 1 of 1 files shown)


nrc:
Thanks for the PR! I think a better solution would be to add a CWD option to the settings and only use that if set, rather than changing the default. Would that solve your use case?

joepie91:
I'd be fine with having this be an extension option (so long as the CWD is based on the active project/crate, not hardcoded in the options), though I can't see how changing this default would break any existing usecases, especially since the equivalent Atom extension has been implementing this CWD behaviour for quite some time without any issues that I'm aware of.

joepie91:
Hi, any update on this?

Xanewok:
See https://github.com/rust-lang/rls-vscode/pull/514/files#r266254468

bkchr:
@joepie91 will you update your pr or should I update my pr? :D

joepie91:
@bkchr I'm a little busy at the moment and don't have a lot of time to track my PRs, so it would probably be best if you updated your PR instead :)

bkchr:
Okay done, maybe you could then close this pr :)

Xanewok:
Done in #514. Thank you @joepie91 and @bkchr!

PR #469, opened 2018-12-08
Change Uri.parse to Uri.file

### LINKED ISSUE #468: Panic with RLS 1.31.0 and latest vscode extension
RLS is panicking on 1.31.0 with vscode extension v0.5.1 whenever opening a folder that uses cargo workspaces.  Opening any sub-folder works.  v0.5.2 looks like it was intended to address this based on the changelog, but I am still getting the issue under that version.

Since RLS and the VSCode extension were updated at around the same time, it's not clear which is causing the issue.  However it seems very-specific to vscode.  See this issue: https://github.com/rust-lang/rls/issues/1173

crlf0710:
Fixes #468

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -117,7 +117,7 @@ function getCargoTomlWorkspace(cur_workspace: WorkspaceFolder, file_path: string
 
         const cargo_path = path.join(current, 'Cargo.toml');
         if (fs.existsSync(cargo_path)) {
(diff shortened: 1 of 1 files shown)


crlf0710:
The relevant code is rolled back in 0.5.2, i think. This was an attempt to fix an issue in those removed code. I'm not sure if there's plan for those code coming back.

Xanewok:
Merged as a part of https://github.com/rust-lang/rls-vscode/pull/759

PR #474, opened 2018-12-16
Recommend new VS Code TSLint plugin

tchebb:
From the description of eg2.tslint: "Note: We are about to deprecate this extension in favor of the vscode-typescript-tslint-plugin."

Update extensions.json in .vscode to respect this.


nrc:
Thanks!

PR #481, opened 2018-12-20
Update link in README.md

kassens:
The link is no longer available, this updates it to the new location.

An alternative link would be versioned to 1.30: https://doc.rust-lang.org/1.30.0/book/second-edition/ch01-03-hello-cargo.html


nrc:
Thanks!

PR #483, opened 2018-12-21
Updating and adding more snippets

YakoYakoYokuYoku:
Added function, parametrized function, while, while let, match snippets.

### DIFF
--- a/snippets/rust.json
+++ b/snippets/rust.json
@@ -15,6 +15,24 @@
         ],
         "description": "Insert for loop"
     },
+    "while": {
+        "prefix": "while",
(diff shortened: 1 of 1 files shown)


YakoYakoYokuYoku:
Please don't ignore me :pray:

nrc:
Hi, thanks for the PR! Sorry for the delay in review - I've been on vacation. I prefer to limit snippets to a small number and only for complex use cases, we recommend users install Trusty Rusty Snippets if they want a more complete set of snippets. As such, these might be better submitted there if they are missing from that extension.

PR #486, opened 2019-01-04
Add an option for rust.trace.server.

bolinfest:
Because this extension leverages `vscode-languageclient`,
it can provide nice logging support for the RLS "for free"
by exposing a setting named `rust.trace.server`.
See https://code.visualstudio.com/api/language-extensions/language-server-extension-guide#logging-support-for-language-server
for details.

### DIFF
--- a/package.json
+++ b/package.json
@@ -409,7 +409,7 @@
                         "verbose"
                     ],
                     "default": "off",
-                    "description": "Traces the communication between VS Code and the RLS.",
+                    "description": "Traces the communication between VS Code and the Rust language server.",
                     "scope": "window"
                 }
             }
--- a/package.json
+++ b/package.json
@@ -400,6 +400,17 @@
                     "default": true,
                     "description": "Show additional context in hover tooltips when available. This is often the type local variable declaration.",
                     "scope": "resource"
+                },
+                "rust.trace.server": {
+                    "type": "string",
+                    "enum": [
+                        "off",
+                        "messages",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -400,6 +400,17 @@
...
+                },
+                "rust.trace.server": {
+                    "type": "string",
+                    "enum": [
+                        "off",
+                        "messages",
+                        "verbose"
+                    ],
+                    "default": "off",
+                    "description": "Traces the communication between VS Code and the RLS.",
```
Xanewok: Minor nit:
```suggestion
                    "description": "Traces the communication between VS Code and the Rust language server.",
```
To keep parity with some of the existing, built-in language servers:
rust trace

src/extension.ts
```diff
@@ -221,7 +221,7 @@ class ClientWorkspace {
         };
 
         // Create the language client and start the client.
-        this.lc = new LanguageClient('Rust Language Server', serverOptions, clientOptions);
+        this.lc = new LanguageClient('rust', 'Rust Language Server', serverOptions, clientOptions);
```
Xanewok: I feel stupid now :man_facepalming: I remember trying to make `[langId].trace.server` work but couldn't. So that was the missing bit!

bolinfest: If it makes you feel any better, I had to step through with the debugger to figure this out!

Xanewok:
Thank you!

PR #497, opened 2019-01-28
Allow to not look for config in outermost nested folder.

### LINKED ISSUE #495: Multi-root workspaces: handling of nested roots (getOuterMostWorkspaceFolder)
TL:DR It seems that replacing workspace folders with outermost folder in `extension.ts` (`getOuterMostWorkspaceFolder(folder)`) breaks the multi-root workspace support for the case when one root is nested in another AND it's the nested one that is a Rust project. Could the use of  `getOuterMostWorkspaceFolder(folder)` be configurable? (I realize it's needed for the opposite case – when one root is nested in another and it's the outermost one that is a Rust project / has cargo.toml.)

Longer version:

I'm working in a codebase where the top-level project is not rust, but it contains rust projects in subdirs. My root folders are:

```
{
	"folders": [
		{
			"path": "~/projects/fuchsia/path/to/subproject"
		},
		{
			"path": "~/projects/fuchsia"
		}
	],
	"settings": {}
}
```
~/projects/fuchsia/path/to/subproject is a Rust project and has Cargo.toml
~/projects/fuchsia is not a Rust project and doesn't have Cargo.toml

It seems that rls-vscode almost works for this setup, but not quite – because when opening a file under subproject, we intentionally not look for the cargo file ~/projects/fuchsia/path/to/subproject , but rather in ~/projects/fuchsia , due to use of `getOuterMostWorkspaceFolder(folder)`. Could this please be configurable?

ppiet:
This patch makes it configurable where the extension looks for Rust
config relevant to a file in case of nested top-level directories.

Example: imagine a multi-language project in `<project_path>` with a
Rust component under `<project_path>/<rust subdir>`. In this case a
developer might open a workspace that contains both `<project_path>` and
`<project_path>/<rust subdir>` as two different top-level directories
(the former for easy access to the Rust code they work on, the latter to
occasionally look up interfaces and docs from outside of the Rust
directory.

This setup doesn't work without this patch: the extension is looking for
Rust config always in the outermost of the possibly multiple workspace
folders that contain the file.

This patch makes the behavior configurable, so that for the setup above
one can set:

  `"rust-client.nestedMultiRootConfigInOutermost": false`

and have the extension working in `<rust subdir>` despite also having
`<project path>` open.

Fixes #495.

### DIFF
--- a/package.json
+++ b/package.json
@@ -216,6 +216,11 @@
                     "default": null,
                     "description": "Rust channel to invoke rustup with. Ignored if rustup is disabled. By default, uses the same channel as your currently open project."
                 },
+                "rust-client.nestedMultiRootConfigInOutermost": {
+                    "type": "boolean",
+                    "default": true,
+                    "description": "If one root workspace folder is nested in another root folder, look for the Rust config in the outermost root."
+                },
                 "rust.sysroot": {
                     "type": [
                         "string",
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -57,7 +57,9 @@ function didOpenTextDocument(document: TextDocument, context: ExtensionContext):
     if (!folder) {
         return;
     }
-    folder = getOuterMostWorkspaceFolder(folder);
+    if (workspace.getConfiguration().get<boolean>('rust-client.nestedMultiRootConfigInOutermost', true)) {
+        folder = getOuterMostWorkspaceFolder(folder);
+    }
     // folder = getCargoTomlWorkspace(folder, document.uri.fsPath);
     if (!folder) {
         stopSpinner(`RLS: Cargo.toml missing`);


ppiet:
Any thoughts on this pull request? Would you consider merging it in? Is there anything I can do to improve it?

ppiet:
Abandoning due to lack of feedback. If anybody would like to pick it up, feel free!

Xanewok:
Sorry for the radio silence. I'd be inclined to merge this, however I think project layout detection should be improved in general (I believe that'd be mostly on the RLS side), so I'd like to think of it as a temporary solution. However, if it solves a real use case then I'd be happy to go with it!

qm3ster:
@ppiet can you reopen this? It doesn't look like it went stale, and I would like this feature.

PR #514, opened 2019-02-27
Set cwd to workspace root path

bkchr:


bkchr:
@Xanewok could you take a look at this?

src/extension.ts
```diff
@@ -437,11 +437,15 @@ class ClientWorkspace {
         // We don't need to set [DY]LD_LIBRARY_PATH if we're using rustup,
         // as rustup will set it for us when it chooses a toolchain.
         const env = await this.makeRlsEnv(/*setLibPath*/ this.config.rustupDisabled);
+        let cwd = undefined;
+        if (workspace.workspaceFolders) {
+            cwd = workspace.workspaceFolders[0].uri.fsPath;
```
Xanewok: IIUC we shouldn't prioritise the first workspace folder but just use the one associated with a given RLS invocation:

https://github.com/rust-lang/rls-vscode/blob/04f5ff8aa270bed9e3b8ec00a6a21a9fc42bd3ee/src/extension.ts#L182

Could you change it to use that, instead?

Xanewok: I believe https://github.com/rust-lang/rls-vscode/pull/459/files#diff-45327f86d4438556066de133327f4ca2R374 is exactly what we want (sorry @joepie91!) - I'd be fine with either this line changed to match the https://github.com/rust-lang/rls-vscode/pull/459 or that PR but rebased

(Really sorry about the delay, the RLS extension didn't get much love recently)

bkchr:
@Xanewok done.

Xanewok:
Thank you!

PR #518, opened 2019-03-07
Tiny change to README.md

angusgraham:
The command `ext install rust` doesn't work in the command palette you get by pressing ctrl + shift + p or by selecting "command palette" from the menus (gives No commands matching), so clarify that you need to press (ctrl + p).
The 'run tasks' command is actually 'run task'


Xanewok:
Looks good, thanks!

Issue #517, opened 2019-03-07
RLS could not be started


greg5678:
I just installed rustup and VSCode on MacOS Sierra today and when I use cargo to compile rust files it works fine but VSCode is giving me issues. 

`Couldn't start client Rust Language Server
`
`Unexpected error initialising RLS - error running rustup
`
`RLS could not set RUST_SRC_PATH for Racer because it could not read the Rust sysroot.
`

JMS55:
Same exact issue. Rls was working fine, then I upgraded to 1.33.0 and I get the same error.

davidarmstronglewis:
Same issue on macOS Mojave

```
stable-x86_64-apple-darwin (default)
beta-x86_64-apple-darwin
nightly-x86_64-apple-darwin

rustc --version
rustc 1.33.0 (2aa4c46cf 2019-02-28)
```

rbudnar:
I also just encountered this on Windows 10 (I'm running on nightly). Here's a snapshot of the errors in the VS code devtools in case it helps:
![image](https://user-images.githubusercontent.com/1072625/53928818-204da980-4051-11e9-9f19-9d53b153ff99.png)

boardwalk:
Thanks @rbudnar, that worked for me ("nightly-x86_64-apple-darwin" in my case).

DanielJoyce:
Pooping the bed for me too on Linux. Will this ever become stable?

I also like when it crashes, but you still have to manually restart. Can't the plugin watch the stdout for error messages and just handle it?

DanielJoyce:
I thought we were done with having to use nightly channels...

rbudnar:
Note you should be able to use stable if you choose. I just so happen to use nightly, so that's what I pasted in there.

colinwd:
Can confirm, stable channel works fine. The issue is that the automatic toolchain name parsing is including `(default)` in the command to `rustup` when it shouldn't be.

massivetim:
I have the same issue. Thanks @rbudnar for the solution.

kinnison:
This is related to `rustup` version 1.17.0 changing its UI for improving things for humans.

bashfulrobot:
> I think you should be able to use stable if you choose. I just so happen to use nightly, so that's what I pasted in there as an example.

Confirmed. 

I literally had JUST installed Rust to start out and ran into this.

* Linux
* Latest VSCode

[...]

PR #520, opened 2019-03-07
rustup.ts: Update to cope with rustup 1.17.0

kinnison:
Rustup 1.17.0 introduced some extra information in the output of
`rustup show active-toolchain` which was confusing the channel
selector.  This patch deals with that.

Fixed: #517

Signed-off-by: Daniel Silverstone <dsilvers@digital-scurf.org>


kinnison:
Also, I'd suggest that if this *is* acceptable and enough, that a release happens pretty soon or we're just going to get more and more reports of an inability to use rls-vscode as people upgrade rustup

Xanewok:
Looks good. Thanks! Will try to do a release soon.

SimonSapin:
Confirmed fixed in 0.5.4. Thanks!

PR #523, opened 2019-03-10
Change `ext install rust` to  `ext install rust-lang.rust`

angusgraham:
So that when you press enter it installs the extension instead of starting a search in the extension browser for all extensions that contain `rust`


Xanewok:
Thank you!

PR #528, opened 2019-03-12
Show more details of what went wrong during the installation of RLS

ricardopieper:
I'm getting a specific error (the rustup add component rls command itself is not running for whatever reason) that I only discovered after the changes I did in this pull request. 

This basically shows more details of the error in the error message popup shown by ` window.showErrorMessage`. It checks if a message exists in the exception and adds it to the text.

I think this is necessary so we can debug the error ourselves. I don't know why the error itself is happening, but the cause seems unrelated to the extension.

I don't know much typescript but there it is.

### DIFF
--- a/src/rustup.ts
+++ b/src/rustup.ts
@@ -148,9 +148,13 @@ async function installRls(config: RustupConfig): Promise<void> {
             console.log(stderr);
             return null;
         }
-        catch (_e) {
-            window.showErrorMessage(`Could not install RLS component (${component})`);
-            const err = new Error(`installing ${component} failed`);
+        catch (e) {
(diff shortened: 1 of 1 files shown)


ricardopieper:
The error happened because today's nightly has no `rls`. Fixed on my end since I need to debug some macros.

So far I've seen 2 issues blocking this extension from not "just working": nightly builds and rustup overrides. The first I believe can be solved by switching to the stable build by default, and nightly via opt-in.

The second one I'm not sure. Perhaps informing the user that RLS couldn't be started and there's an override in place, and offer to remove it and try again?

Xanewok:
Makes sense, thank you!

PR #538, opened 2019-03-21
Enable logging on Windows

SoftwareApe:
Before the code used path, which yields an illegal path name on Windows (starting with /C:/...). fsPath gives the proper pathname.

This change should be tested on Linux before it is merged. I don't have a Linux available right now to test it.


Xanewok:
Nice catch, thanks!

PR #539, opened 2019-03-22
Patch for 111e12c by renaming language id from rust to rust-client.

TheGoddessInari:
The language id seems to be simply an identifier and unused elsewhere.

Fixes rust-lang/rls-vscode#530.

It's tested and appears to be working.


Xanewok:
Yeah, the patch looks good. Thank you!

PR #543, opened 2019-03-22
Clean up (part 3)

Xanewok:
Things left to do:
- Finish moving setup (installing toolchain etc.) to task-based workflow (UX is a lot better)
- Separate LSP/toolchain setup to different files

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -112,10 +112,10 @@ export class RLSConfiguration {
 
   // Added ignoreChannel for readChannel function. Otherwise we end in an infinite loop.
   public rustupConfig(ignoreChannel: boolean = false): RustupConfig {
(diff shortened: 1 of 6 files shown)


Xanewok:
@yisonPylkita if possible, could I ask you to test if WSL still works as expected with this merged change? I was extra careful not to mess anything but something might've slipped by - I'd love to catch that before releasing a new version. I'd be much obliged!

yisonPylkita:
Wow, That was quite a refactor!
I will check `master` branch tomorrow on WSL

yisonPylkita:
I did some tests on `master` and it looks good (beside `rls` still missing in lastest nightly)

Xanewok:
Awesome, thank you for that!

Issue #199, opened 2017-11-27
Feature request: add tooltip for method arguments


peperunas:
Hello!

It would be great to have a tooltip which helps you remember which argument has to be passed to a method.

This is what I mean (taken from a `.c` source file):

![Tooltip example](https://user-images.githubusercontent.com/6033387/33278356-960a8d42-d39b-11e7-9ff3-304d1352ab3c.png)

Keep up the good work!

nrc:
We already have this information - we display it on hover of the function name, so we'd just need to surface that as a tooltip. I assume there is an LSP message for that, though I don't know what that is.

jonasbb:
I think this is tracked as https://github.com/rust-lang-nursery/rls/issues/97I don't think it is an easy issue. 

The LSP message would be
https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#textDocument_signatureHelp

nrc:
I was thinking there might be an easier solution than full signature help - just pop up a tooltip quoting the function signature when the cursor is in function parens. Not sure if that is possible, but if it is, then it would be much easier.

viktorstrate:
In Javascript and other languages, I can press <kbd>⌘</kbd>+<kbd>Shift</kbd>+<kbd>Space</kbd> to call the `Trigger Parameter Hints` command, which I find extremely useful. But it doesn't work with this plugin.

The shortcut is only activated when `editorHasSignatureHelpProvider` is true.

Is there any updates on this issue?

radu-matei:
If anyone is interested in this feature, have a look at #552.
Feedback appreciated, thanks!

jens1o:
Thanks for your work!
I wonder whether it would be possible to syntax highlight the signature help, like it is when hovering the method?

radu-matei:
@jens1o -- Unfortunately, the type of the label in the signature help object is a plain string, so we have to strip the markdown formatting of the function signature, meaning it's just plain text (this is how all other extensions that add VS Code language support display the signature help label).

jens1o:
Okay, that's disappointing, but thanks for your work! :)

radu-matei:
If VS Code gets an update on the label, we can also change it -- but otherwise, I don't think there's much we can do here about the formatting.

That being said, there are things we can improve -- around highlighting the current parameter -- but for this, we need to use the signature help request with the RLS, which is not currently implemented.

[...]

PR #552, opened 2019-03-28
Add initial tooltip for function signature provider

radu-matei:
This PR addresses #199.

Since RLS doesn't have support for signature help provider, this PR uses the hover provider implemented by RLS to get some basic information about the function signature and display it as a tooltip when calling functions -- see demo below.

How this works:

- it uses the `registerSignatureHelpProvider` from the VS Code API, triggered by `(` and `,`
- it then creates a hover request to RLS using the entity right before `(` and constructs a `vscode.SignatureHelp` object, which is displayed as a tooltip (@nrc's approach, suggested in #199).
- because there is no information about the number of type of characters from RLS (and implementing this in JS means either writing a basic parser or a regex), in its current form, the tooltip is shown  whenever the `,` character is used and `context.isRetrigger` is true -- which corresponds to passing a list of arguments to a function. This way, we make sure the tooltip is not triggered when writing an enumeration, for example. (Note that there should be some more testing around this, to make sure the tooltip isn't triggered in random places where commas are used.)

- also because there is no information about the number of characters, we cannot write the current parameter in bold (I guess we could, but it would require some serious string manipulation for the signature, as well as temporarily recording user input and counting commas.)

- various decisions and assumptions are described in the comments, and I'd be more than happy to discuss around them.

Thanks!

rls-vscode-gif

### DIFF
--- a/src/providers/signatureHelpProvider.ts
+++ b/src/providers/signatureHelpProvider.ts
@@ -1,0 +1,155 @@
+import * as vscode from 'vscode';
+import { HoverRequest, LanguageClient } from 'vscode-languageclient';
+
+export class SignatureHelpProvider implements vscode.SignatureHelpProvider {
+  private languageClient: LanguageClient;
+  private previousFunctionPosition?: vscode.Position;
+
+  constructor(lc: LanguageClient) {
+    this.languageClient = lc;
+  }
+
+  public provideSignatureHelp(
+    document: vscode.TextDocument,
+    position: vscode.Position,
+    token: vscode.CancellationToken,
+    context: vscode.SignatureHelpContext,
+  ): vscode.ProviderResult<vscode.SignatureHelp> {
+    // the current signature help provider uses the hover information from RLS
+    // and it only has a string representation of the function signature.
+    // This check makes sure we can easily show the tooltip for multiple parameters, separated by `,`
+    if (context.triggerCharacter === '(') {
+      this.previousFunctionPosition = position;
+      return this.provideHover(
+        this.languageClient,
+        document,
+        position,
+        token,
+      ).then(hover => {
+        return this.hoverToSignatureHelp(hover);
+      });
+    } else if (context.triggerCharacter === ',') {
+      if (this.previousFunctionPosition) {
+        return this.provideHover(
+          this.languageClient,
+          document,
+          this.previousFunctionPosition,
+          token,
+        ).then(hover => {
+          return this.hoverToSignatureHelp(hover);
+        });
+      } else {
+        return null;
+      }
+    } else {
+      if (context.isRetrigger === false) {
+        this.previousFunctionPosition = undefined;
+      }
+      return null;
+    }
+  }
+
+  private provideHover(
+    lc: LanguageClient,
+    document: vscode.TextDocument,
+    position: vscode.Position,
+    token: vscode.CancellationToken,
+  ): Promise<vscode.Hover> {
+    return new Promise((resolve, reject) => {
+      lc.sendRequest(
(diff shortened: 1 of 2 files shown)


src/providers/signatureHelpProvider.ts
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  private provideHover(
+    lc: LanguageClient,
+    document: vscode.TextDocument,
+    position: vscode.Position,
+    token: vscode.CancellationToken,
+  ): Promise<vscode.Hover> {
+    return new Promise((resolve, reject) => {
+      lc.sendRequest(
```
Xanewok: I know that VSCode mostly uses a `Thenable<T>` interface but does it support `await`ing?
If so, I think it'd be good to
```ts
const response = await lc.sendRequest(...);
return lc.protocol2CodeConverter.asHover(response);
```
(assuming it'd throw an error correctly)

src/providers/signatureHelpProvider.ts
```diff
@@ -0,0 +1,155 @@
...
+      .replace('```rust', '')
+      .replace('```', '');
+
+    // the signature help tooltip is activated on `(` or `,`
+    // and without this, it could show the tooltip after non-functions
+    if (!label.includes('fn')) {
+      return undefined;
+    }
+
+    let doc: vscode.MarkdownString | undefined;
```
Xanewok: I think we could simplify to `const doc = hover.contents.length > 1 ? hover.contents.slice(-1) : undefined;` I personally find it more readable and also concise

src/providers/signatureHelpProvider.ts
```diff
@@ -0,0 +1,155 @@
...
+    // This check makes sure we can easily show the tooltip for multiple parameters, separated by `,`
+    if (context.triggerCharacter === '(') {
+      this.previousFunctionPosition = position;
+      return this.provideHover(
+        this.languageClient,
+        document,
+        position,
+        token,
+      ).then(hover => {
+        return this.hoverToSignatureHelp(hover);
```
Xanewok: For such simple functions I'd prefer to use the `() => returnValue` shorthand. I think we should enable https://palantir.github.io/tslint/rules/arrow-return-shorthand/ rule in some form or another later on but let's not worry about that now :+1:

src/providers/signatureHelpProvider.ts
radu-matei: If we were to `await` here, the containing function, `provideSignatureHelp`, would need to change its signature and become `async` -- which I think means it would no longer correctly implement the `vscode.SignatureHelpProvider` interface.

src/providers/signatureHelpProvider.ts
radu-matei: I think it would have to be `const doc = hover.contents.length > 1 ? hover.contents.slice(-1)[0] : undefined;`, but yes, it is more readable indeed.

src/providers/signatureHelpProvider.ts
Xanewok: Argh, that might be the case, you’re right. It’s unfortunate the API is settled on the lowest promise common denominator at the time but there’s probably nothing we can do at the time cleanly

src/providers/signatureHelpProvider.ts
radu-matei: Definitely more readable this way, updating.

radu-matei:
@Xanewok -- addressed feedback, plus added another check to ensure we don't show the tooltip for `,` outside of function calls.

Thanks!

Xanewok:
That's great, thanks again!

PR #553, opened 2019-03-29
Special characters handling in WSL paths

yisonPylkita:
WSL will not work for projects with special characters in path.
This PR aims to fix this.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -258,6 +258,27 @@ class ClientWorkspace {
             wslPath = wslPath.substr('file://'.length);
           }
 
+          wslPath = wslPath.split('%20').join(' ');
+          wslPath = wslPath.split('%21').join('!');
+          wslPath = wslPath.split('%23').join('#');
+          wslPath = wslPath.split('%24').join('$');
+          wslPath = wslPath.split('%26').join('&');
+          wslPath = wslPath.split('%27').join('\'');
+          wslPath = wslPath.split('%28').join('(');
+          wslPath = wslPath.split('%29').join(')');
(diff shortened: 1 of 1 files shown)


yisonPylkita:
Fix for https://github.com/rust-lang/rls-vscode/issues/550

Xanewok:
Could we use a more complete approach like https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURI instead of special-casing these?

Xanewok:
@yisonPylkita by the way, will `wslPath` always be an URI? If so, I think we should use `URI.parse` and use the `URI.parse(wslPath).fsPath` for the conversion

src/extension.ts
```diff
@@ -258,6 +258,27 @@ class ClientWorkspace {
...
+          wslPath = wslPath.split('%2D').join('-');
+          wslPath = wslPath.split('%2E').join(',');
+          wslPath = wslPath.split('%3D').join('=');
+          wslPath = wslPath.split('%40').join('@');
+          wslPath = wslPath.split('%5B').join('[');
+          wslPath = wslPath.split('%5D').join(']');
+          wslPath = wslPath.split('%5E').join('^');
+          wslPath = wslPath.split('%7B').join('{');
+          wslPath = wslPath.split('%7D').join('}');
+          wslPath = wslPath.split('%7E').join('~');
```
Xanewok: For future reference, while this looks declarative (although it's not immediately obvious it's a replace operation), this might not be good from the performance standpoint. Always profile first, but doing this many string manipulations (and separately, one after another) looks like an unnecessary performance overhead

PR #554, opened 2019-03-29
Introduce extension tests

Xanewok:
This uses most of the setup from https://code.visualstudio.com/api/working-with-extensions/testing-extension and so required some tweaks in `tsconfig.json`.

We now test on 3 platforms and also can test extension-specific behaviour, which are effectively integration-style tests.


Xanewok:
The PR is is green on Travis, no idea why there are duplicate checks here (one stuck in the queued state).

src/test/utils/wslpath.test.ts
```diff
@@ -0,0 +1,33 @@
+import * as assert from 'assert';
+
+import * as wslpath from '../../utils/wslpath';
+
+function mkConverterCheck(converter: (arg: string) => string) {
+  return (from: string, to: string) => assert.equal(converter(from), to);
+}
+
+suite('WSL path conversion', () => {
```
Xanewok: @yisonPylkita seeing as you're more knowledgeable about the WSL path conversion, could you please add more tests to better cover the existing/desired behaviour? (It'd also might be a good idea to include tests when making changes like https://github.com/rust-lang/rls-vscode/issues/550)

PR #567, opened 2019-04-10
Use wslpath to do win-wsl path transform

LightMelodies:
The orign code in _wslpath.ts_ will failed in some situation, e.g. the drvfs mount point is not in _/mnt_. So I change it to use wslpath to do the path transform.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -15,7 +15,7 @@
 			"outFiles": [
 				"${workspaceFolder}/out/**/*.js"
 			],
-			"preLaunchTask": "npm: watch"
+			"preLaunchTask": "watch"
 		},
 		{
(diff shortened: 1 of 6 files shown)


Xanewok:
The reason we did this in-extension was because of latency - each call to wslpwth generated 50-100ms worth of latency, which is considerable and I’d be willing to fix the issues with the existing approach, but don’t shell out to wslpath more than once, e.g. to detect mount points.

cc @yisonPylkita who implemented the path translation

Xanewok:
In addition to the comment above, VSCode started supporting WSL via a dedicated extension https://code.visualstudio.com/docs/remote/wsl so I'm not sure if we'll continue supporting this directly due to the problems with path translation.

PR #568, opened 2019-04-10
customize wsl mount point

LightMelodies:
Allow customize wsl mount point via **rust-client.wslMountPoint.**, with some fix in path translation and cargo task.

### DIFF
--- a/src/tasks.ts
+++ b/src/tasks.ts
@@ -10,6 +10,8 @@ import {
   WorkspaceFolder,
 } from 'vscode';
 
+import { RLSConfiguration } from './configuration';
+
 /**
  * Displayed identifier associated with each task.
  */
@@ -42,7 +44,9 @@ export interface Execution {
  */
 function createShellExecution(execution: Execution): ShellExecution {
   const { binary, command, args, cwd, env } = execution;
-  const cmdLine = `${command || binary} ${args.join(' ')}`;
+  const cmdLine = RLSConfiguration.loadFromWorkspace('').useWSL
+    ? `wsl.exe ${command || binary} ${args.join(' ')}`
+    : `${command || binary} ${args.join(' ')}`;
   return new ShellExecution(cmdLine, { cwd, env });
 }
 
--- a/src/test/utils/wslpath.test.ts
+++ b/src/test/utils/wslpath.test.ts
@@ -2,32 +2,32 @@ import * as assert from 'assert';
 
 import * as wslpath from '../../utils/wslpath';
 
-function mkConverterCheck(converter: (arg: string) => string) {
-  return (from: string, to: string) => assert.equal(converter(from), to);
+function mkConverterCheck(converter: (arg: string, m: string) => string) {
+  return (from: string, to: string) => assert.equal(converter(from, '/'), to);
 }
 
 suite('WSL path conversion', () => {
   test('uriWindowsToWsl', () => {
     const check = mkConverterCheck(wslpath.uriWindowsToWsl);
     // Basic test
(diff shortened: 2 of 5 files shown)


src/tasks.ts
```diff
@@ -42,7 +44,9 @@ export interface Execution {
  */
 function createShellExecution(execution: Execution): ShellExecution {
   const { binary, command, args, cwd, env } = execution;
-  const cmdLine = `${command || binary} ${args.join(' ')}`;
+  const cmdLine = RLSConfiguration.loadFromWorkspace('').useWSL
```
Xanewok: At a first glance it'd be better to modify the `Execution` before passing it to this function or introduce a `useWsl: boolean` argument here and propagate it downwards as needed

src/test/utils/wslpath.test.ts
```diff
@@ -2,32 +2,32 @@ import * as assert from 'assert';
 
 import * as wslpath from '../../utils/wslpath';
 
-function mkConverterCheck(converter: (arg: string) => string) {
-  return (from: string, to: string) => assert.equal(converter(from), to);
+function mkConverterCheck(converter: (arg: string, m: string) => string) {
+  return (from: string, to: string) => assert.equal(converter(from, '/'), to);
```
Xanewok:
```suggestion
  return (from: string, to: string) => assert.equal(converter(from, mnt), to);
```

src/test/utils/wslpath.test.ts
```diff
@@ -2,32 +2,32 @@ import * as assert from 'assert';
 
 import * as wslpath from '../../utils/wslpath';
 
-function mkConverterCheck(converter: (arg: string) => string) {
-  return (from: string, to: string) => assert.equal(converter(from), to);
+function mkConverterCheck(converter: (arg: string, m: string) => string) {
```
Xanewok:
```suggestion
function mkConverterCheck(converter: (arg: string, mnt: string) => string) {
```

src/test/utils/wslpath.test.ts
```diff
@@ -2,32 +2,32 @@ import * as assert from 'assert';
...
+  return (from: string, to: string) => assert.equal(converter(from, '/'), to);
 }
 
 suite('WSL path conversion', () => {
   test('uriWindowsToWsl', () => {
     const check = mkConverterCheck(wslpath.uriWindowsToWsl);
```
Xanewok: With the argument changes, we'd need to pass the mount point parameter:
```typescript
const check = mkConverterCheck(wslpath.uriWindowsToWsl, '/mnt');
```

src/test/utils/wslpath.test.ts
```diff
@@ -2,32 +2,32 @@ import * as assert from 'assert';
...
-    check('C:\\Program Files\\somedir\\', '/mnt/c/Program Files/somedir/');
+    check('C:\\Program Files\\somedir\\', '/c/Program Files/somedir/');
     // Different disk letter
-    check('z:\\', '/mnt/z/');
-    check('C:\\', '/mnt/c/');
+    check('z:\\', '/z/');
+    check('C:\\', '/c/');
   });
   test('uriWslToWindows', () => {
     const check = mkConverterCheck(wslpath.uriWslToWindows);
```
Xanewok: Same with https://github.com/rust-lang/rls-vscode/pull/568/files#r274384815

Xanewok:
WSL support is now deprecated so I'll go ahead and close this PR. Thanks for the PR - hopefully the official Remote - WSL should give us better support than trying to hack around it.

PR #581, opened 2019-04-23
Clarify how clippy_preference works

nickolay:
Syncs the extension's docs with the RLS readme (which was updated in https://github.com/rust-lang/rls/pull/1439 per discussion in https://github.com/rust-lang/rls/issues/1437 )


Xanewok:
Thanks!

PR #638, opened 2019-08-07
reintroduce multi project setup

jannickj:
This will allow you to have project structures like this:
```
projectA/Cargo.toml
projectA/src
projectB/Cargo.toml
projectB/src
subProjects/projectD/Cargo.toml
subProjects/projectD/src
subProjects/projectE/Cargo.toml
subProjects/projectE/src
```

when you open a file it will search up parent folders until it finds a Cargo.toml, and either reuse the RLS instance for that one or create one if it doesn't exists.

Changes
* enable multiple projects with multiple rls running
* change the `URI` key back to `string` for` workspaces` (as it's unreliable)
* add selectors such that only the correct rls is used for a given project
* take great care to ensure the old setup works as normal
* add a configuration and turn it off by default.

Second attempt at: #457

### DIFF
--- a/package.json
+++ b/package.json
@@ -219,6 +219,11 @@
                     "default": true,
                     "description": "If one root workspace folder is nested in another root folder, look for the Rust config in the outermost root."
                 },
+                "rust-client.enableMultiProjectSetup": {
+                    "type": "boolean",
+                    "default": false,
+                    "description": "Allow multiple projects in the same folder, along with remove the constraint that the cargo.toml must be located at the root. (Experimental: might not work for certain setups)"
+                },
                 "rust.sysroot": {
                     "type": [
                         "string",
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -69,23 +69,37 @@ function didOpenTextDocument(
   if (!folder) {
     return;
   }
+
   if (
+    workspace
+      .getConfiguration()
+      .get<boolean>('rust-client.enableMultiProjectSetup', false)
+  ) {
+    folder = getCargoTomlWorkspace(folder, document.uri.fsPath);
+  } else if (
     workspace
       .getConfiguration()
       .get<boolean>('rust-client.nestedMultiRootConfigInOutermost', true)
   ) {
     folder = getOuterMostWorkspaceFolder(folder);
   }
-  // folder = getCargoTomlWorkspace(folder, document.uri.fsPath);
+
   if (!folder) {
     stopSpinner(`RLS: Cargo.toml missing`);
     return;
   }
 
-  if (!workspaces.has(folder.uri)) {
+  const folderPath = folder.uri.toString();
+
+  if (!workspaces.has(folderPath)) {
+
     const workspace = new ClientWorkspace(folder);
-    workspaces.set(folder.uri, workspace);
+    activeWorkspace = workspace;
+    workspaces.set(folderPath, workspace);
     workspace.start(context);
+  } else {
+    const ws = workspaces.get(folderPath);
+    activeWorkspace = typeof ws === "undefined" ? null : ws;
   }
 }
 
@@ -110,37 +124,37 @@ function sortedWorkspaceFolders(): string[] {
   return _sortedWorkspaceFolders || [];
 }
 
-// function getCargoTomlWorkspace(cur_workspace: WorkspaceFolder, file_path: string): WorkspaceFolder {
-//     if (!cur_workspace) {
-//         return cur_workspace;
-//     }
+function getCargoTomlWorkspace(cur_workspace: WorkspaceFolder, file_path: string): WorkspaceFolder {
+    if (!cur_workspace) {
+        return cur_workspace;
+    }
 
-//     const workspace_root = path.parse(cur_workspace.uri.fsPath).dir;
-//     const root_manifest = path.join(workspace_root, 'Cargo.toml');
-//     if (fs.existsSync(root_manifest)) {
-//         return cur_workspace;
-//     }
+    const workspace_root = path.parse(cur_workspace.uri.fsPath).dir;
+    const root_manifest = path.join(workspace_root, 'Cargo.toml');
+    if (fs.existsSync(root_manifest)) {
+        return cur_workspace;
+    }
 
-//     let current = file_path;
+    let current = file_path;
 
-//     while (true) {
-//         const old = current;
-//         current = path.dirname(current);
-//         if (old == current) {
-//             break;
-//         }
-//         if (workspace_root == path.parse(current).dir) {
-//             break;
-//         }
+    while (true) {
+        const old = current;
+        current = path.dirname(current);
+        if (old == current) {
+            break;
+        }
+        if (workspace_root == path.parse(current).dir) {
+            break;
+        }
 
-//         const cargo_path = path.join(current, 'Cargo.toml');
-//         if (fs.existsSync(cargo_path)) {
-//             return { ...cur_workspace, uri: Uri.parse(current) };
-//         }
-//     }
+        const cargo_path = path.join(current, 'Cargo.toml');
+        if (fs.existsSync(cargo_path)) {
+            return { ...cur_workspace, uri: Uri.parse(current) };
+        }
+    }
 
-//     return cur_workspace;
-// }
+    return cur_workspace;
+}
 
 function getOuterMostWorkspaceFolder(folder: WorkspaceFolder): WorkspaceFolder {
   const sorted = sortedWorkspaceFolders();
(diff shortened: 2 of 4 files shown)


jannickj:
forced pushed to make it more clear what I am actually changing :)

src/extension.ts
```diff
@@ -279,31 +312,45 @@ class ClientWorkspace {
...
+      return ws.start(context);
     });
     this.disposables.push(restartServer);
 
     this.disposables.push(
-      commands.registerCommand('rls.run', (cmd: Execution) =>
-        runRlsCommand(this.folder, cmd),
-      ),
+      commands.registerCommand('rls.run', (cmd: Execution) => {
+        const ws =
```
MonsieurMan: Reading this took me some second to understand that ws was workspace, also personal preference but I find ternary easier to read like so :

```typescript
const workspace = (multiProjectEnabled && activeWorkspace)
      ? activeWorkspace
      : this;
runRlsCommand(workspace.folder, cmd);
```

I don't know if it's ok with the tslint configs.

src/extension.ts
```diff
@@ -279,31 +312,45 @@ class ClientWorkspace {
...
+          multiProjectEnabled && activeWorkspace ? activeWorkspace : this;
+        return rustupUpdate(ws.config.rustupConfig());
       },
     );
     this.disposables.push(rustupUpdateDisposable);
 
     const restartServer = commands.registerCommand('rls.restart', async () => {
-      await this.stop();
-      return this.start(context);
+      const ws =
```
MonsieurMan: Same comment about ws -> workspace here :slightly_smiling_face:

package.json
```diff
@@ -219,6 +219,11 @@
                     "default": true,
                     "description": "If one root workspace folder is nested in another root folder, look for the Rust config in the outermost root."
                 },
+                "rust-client.enableMultiProjectSetup": {
```
MonsieurMan: I don't know much about the extension, didn't have the time to clone it yet, but wouldn't it be possible to make the extension work like :

- Normal cargo.toml at root of the project is tried
- If it fails the extension tries to determine if it's not a monorepo

I'd like your opinion on this, if it is quite some work I'd happily lend you a hand :slightly_smiling_face:

src/extension.ts
```diff
@@ -69,23 +69,36 @@ function didOpenTextDocument(
   if (!folder) {
     return;
   }
+
   if (
+    workspace
+      .getConfiguration()
```
MonsieurMan: `workspace.getConfiguration` is also called later on, maybe consider storing it into a `const configuration` and reusing the same variable ?

src/extension.ts
```diff
@@ -69,23 +69,36 @@ function didOpenTextDocument(
...
+
+  if (!workspaces.has(folderPath)) {
     const workspace = new ClientWorkspace(folder);
-    workspaces.set(folder.uri, workspace);
+    activeWorkspace = workspace;
+    workspaces.set(folderPath, workspace);
     workspace.start(context);
+  } else {
+    const ws = workspaces.get(folderPath);
+    activeWorkspace = typeof ws === 'undefined' ? null : ws;
```
MonsieurMan:
```typescript
const workspace = workspaces.get(folderPath);
activeWorkspace = (typeof workspace === 'undefined')
    ? null 
    : workspace;
```

src/extension.ts
```diff
@@ -181,15 +197,18 @@ function didChangeWorkspaceFolders(
...
+      workspaces.delete(folder.uri.toString());
       ws.stop();
     }
   }
 }
 
-const workspaces: Map<Uri, ClientWorkspace> = new Map();
+// Don't use URI as it's unreliable the same path might not become the same URI.
+const workspaces: Map<string, ClientWorkspace> = new Map();
+let activeWorkspace: ClientWorkspace | null;
```
MonsieurMan: Typescript has a builtin for nullable variables :
```typescript
let activeWorkspace?: ClientWorkspace;
```

src/extension.ts
```diff
@@ -279,31 +312,45 @@ class ClientWorkspace {
...
+    }
 
+    commandsUnregistered = false;
     const rustupUpdateDisposable = commands.registerCommand(
       'rls.update',
       () => {
-        return rustupUpdate(this.config.rustupConfig());
+        const ws =
+          multiProjectEnabled && activeWorkspace ? activeWorkspace : this;
+        return rustupUpdate(ws.config.rustupConfig());
```
MonsieurMan:
```typescript
const workspace = (multiProjectEnabled && activeWorkspace) 
    ? activeWorkspace 
    : this;
return rustupUpdate(workspace.config.rustupConfig());
```

package.json
jannickj: > Normal cargo.toml at root of the project is tried

With this change the cargo.toml at root should also just work. 

>If it fails the extension tries to determine if it's not a monorepo
How would you detect, that I could imagine some wtf moments where you add a folder and all of the sudden the tooling stops working. I would rather make configuration for stuff like that.

> I'd like your opinion on this, if it is quite some work I'd happily lend you a hand

In general I would estimate that 99% will just always have it turned on.  So mostly this is about not breaking it for the 1%. 

and thx for the offer :)

src/extension.ts
jannickj: ye, the problem is the function was clearly never meant to take configuration arguments. However it would take some re-architecture to make it nice, and I didn't want to push for too many changes.

src/extension.ts
jannickj: the formatter, which is build server checked, decides stuff like that.

jannickj:
@nrc / @Xanewok  do you have time to look it over?

Xanewok:
Yep! Before merging, I'd like to release a 0.6.2 first with a fix for #639

jannickj:
@Xanewok Are you ready to review now?

wyze:
I've built your branch from source and have been using it for over a week without any issues, in both a multi-project setup and single project.

jannickj:
:wave:

jannickj:
> Are URIs unreliable for workspaces because they can be not normalized/canonicalized?

No, I tried with identical URI's but I noticed that there was some caching going on because when I ran the code in debug I could see that the some private field were being set the second time I used a URI for the same string. Which I concluded was why they hash'ed differently.

 I couldn't get it to work on my machine but maybe I was using it wrong, so if someone knows the answer I'm happy to hear it.

src/extension.ts
```diff
@@ -110,37 +123,40 @@ function sortedWorkspaceFolders(): string[] {
...
-
-//         const cargo_path = path.join(current, 'Cargo.toml');
-//         if (fs.existsSync(cargo_path)) {
-//             return { ...cur_workspace, uri: Uri.parse(current) };
-//         }
-//     }
-
-//     return cur_workspace;
-// }
+function getCargoTomlWorkspace(
```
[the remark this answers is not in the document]
jannickj: I'll add some comments. 
However even though the body might look a bit big, I actually kept it like that to make it simplistic, mostly because it's a function that will run fairly often and I didn't want to risk busy loops.

src/extension.ts
```diff
@@ -69,23 +69,36 @@ function didOpenTextDocument(
...
+
   if (!folder) {
     stopSpinner(`RLS: Cargo.toml missing`);
     return;
   }
 
-  if (!workspaces.has(folder.uri)) {
+  const folderPath = folder.uri.toString();
+
+  if (!workspaces.has(folderPath)) {
```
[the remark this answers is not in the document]
jannickj: While I'm not against refactoring some of the code out, I'm afraid that this will delay this PR even more, as it's now doing two things at once.

jannickj:
> I left some comments below - it'd be great to extract the workspace managing bits outside before we delve into details here since it's getting a bit conflated and harder to see which path we should take with the details of the implementation.

It's good comments and I don't disagree, however I feel a requirement of refactoring some of the designs decisions already made in the code base is a bit much. I'm not against refactoring some of the code however I am very afraid I'll spent a good few hours doing that only to have my PR linger for the next 6 months. 

I'll do it but I'll need your word on it not happening.

jannickj:
I looked into it: I started the work on moving to a "workspace" module the problem is that almost 80% of extensions is about the workspace, more than that the code is littered with global state. Now If one must have global state at least keep it contained to one file. As such I ended up just moving my own function, along with writing some TODOs.

With that said I do think it's a good idea to look at this code and do some refactoring however I don't think this is the PR for it.

jannickj:
:wave:

Xanewok:
Changes look good and I won't block this any longer - sorry for the delay!
And many thanks for taking a look at this, I think a lot of people will be happy =)

jli:
@jannickj thanks so much for the contribution!

Everything appears to be working with my nested subdirectory. However, running the "cargo build" task doesn't seem to work. Do you know if this is meant to be supported?

(more details here: https://github.com/rust-lang/rls-vscode/issues/419#issuecomment-612670488 )

Issue #639, opened 2019-08-07
why unkonwn rls configuration use_crate_blacklist 


theone-pang:
open project show "unkonwn rls configuration use_crate_blacklist " tip evry time,why?

greensirius:
I have the same issues here on my vscode, my current rls is 1.38.0

mati865:
It's because of https://github.com/rust-lang/rls/pull/1509
cc @Xanewok

Xanewok:
Yup, will push a new extension version today.

Sorry for the inconvenience.

arlyon:
Any update on this? Still getting this issue

Xanewok:
Sorry for the delay! I'll publish a version when https://github.com/rust-lang/rust/pull/63472 is merged upstream (it looks like it's going to be tomorrow) - with this, the RLS will be able to emit a user friendly error whenever `use_crate_backlist` is still specified.

zuobaoquan:
Workaround for VSCode:   add `"rust.use_crate_blacklist": false` to settings.json and restart the editor.

mati865:
It's fixed in up upcoming nightly release.

Xanewok:
This was fixed in the newest RLS edition and a new version of the extension was published, so this shouldn't pop up anymore.

stefanos82:
Nope, still the same issue, even with 0.6.3

![image](https://user-images.githubusercontent.com/1777466/64479822-2ae62880-d1c5-11e9-832a-f9c712d9439d.png)
P.S.: I've just started learning Rust and installed this extension; it's the very first message I got...nice welcoming :rofl:

mati865:
New RLS wasn't yet merged into rust.

stefanos82:
> This was fixed in the newest RLS edition and a new version of the extension was published, so this shouldn't pop up anymore.

I based my reply on this quote.

AkariAkaori:
Still getting it with `rls 1.39.0 (80a1d34 2019-09-20)` and extension 0.6.3, don't have `use_crate_blacklist` anywhere in my project directory or user settings.json

dashed:
@Xanewok For some reason, I'm receiving this warning for rust projects. I'm using v0.6.3 of rls-vscode 😕

Xanewok:
Please update the extension to the 0.7 version, the Rust toolchain to 1.38 and let me know if the problem still persists, thanks!

stefanos82:
[...]

PR #643, opened 2019-08-11
Fix use_crate_blacklist error

arlyon:
Fixes issue #639

### DIFF
--- a/package.json
+++ b/package.json
@@ -294,12 +294,6 @@
                     "description": "Show warnings.",
                     "scope": "resource"
                 },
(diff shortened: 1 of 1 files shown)


Xanewok:
Thanks for the PR! I don't think this is a correct fix - we should warn the user about deprecated option (which is what https://github.com/rust-lang/rls/pull/1539 does) instead.

As such, I'll close this (see https://github.com/rust-lang/rls-vscode/issues/639#issuecomment-520265962 for more details)

PR #654, opened 2019-09-04
Fix rust-client.channel.type

AlexTugarev:
This PR fixes the type of the `rust-client.channel` config property. 

`null` value leads to invalid schema. Ajv is failing to compile it:
```
schema is invalid: data.properties['rust-client.channel'].type should be equal to one of the allowed values, data.properties['rust-client.channel'].type[1] should be equal to one of the allowed values, data.properties['rust-client.channel'].type should match some schema in anyOf
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -193,7 +193,7 @@
                 "rust-client.channel": {
                     "type": [
                         "string",
-                        null
+                        "null"
                     ],
                     "enum": [
                         "stable",


Xanewok:
Nice catch, thanks!

AlexTugarev:
@Xanewok is there an ETA for [0.6.2](https://github.com/rust-lang/rls-vscode/milestone/4) and would it include this one?

Xanewok:
Released a 0.6.3 with the fixed package.json

Issue #604, opened 2019-06-18
Problems displayed late


niobium0:
Currently the plugin is building all crates and only once they are all attempted it displays the problems in Problems panel. This is too slow. In comparison `cargo watch -x check` (from "cargo-watch" crate) gives nearly instantaneous feedback. Would be great if problems were "streamed" from RLS.

Xanewok:
RLS is on par with regular `cargo check` on changed file - try setting `"rust.all_targets": false` to not build every crate target in the workspace - this should substantially drop the analysis latency.

niobium0:
I'm setting `"rust.all_targets": false`, it doesn't seem to work for me.

PR #670, opened 2019-09-29
change default `wait_to_build` to 300 ms

janriemer:
This changes the default `rust.wait_to_build` time to 300 ms.
When doing changes in Rust source files, smaller times to build makes the overall
editing-experience much better, because the developer gets faster feedback.

Is there a reason, why this value was at 1500 ms (which is pretty high IMO)?

I'd really like to see this PR merged, because I always thought that it is the RLS or Rustc itself, which is slow, but it was just due to this setting.

Thanks!

Related: #604

### DIFF
--- a/package.json
+++ b/package.json
@@ -289,7 +289,7 @@
                 },
                 "rust.wait_to_build": {
                     "type": "number",
-                    "default": 1500,
+                    "default": 300,
                     "description": "Time in milliseconds between receiving a change notification and starting build.",
                     "scope": "resource"
                 },


Xanewok:
Hm, I always thought we defer to interval that's dynamically detected by the RLS. @alexheretic do you remember what was required in terms of setup to use the dynamic detection feature?

@janriemer which RLS (`rls --version`) are you using?

alexheretic:
RLS will figure out the build wait using the most recent build duration. It will even infer 0ms waits when builds are very fast (< 300ms).

However, if the language server client sends a `wait_to_build` config value this will be used instead, overriding the dynamic wait calculation.

janriemer:
@Xanewok I am using rls 1.38.0 (7b0a20b 2019-08-11).
@alexheretic so for me, the VSCode client has overriden that dynamic wait calculation!? Why? I am seriously wondering whether this is a bug or feature now. I am a bit confused, sorry. :confused: 
If it is a feature, I would like to propose to decrease the `wait_to_build`, because [as you have already seen in another issue](https://github.com/rust-lang/rls/issues/1568), I am not the only one experiencing this issue.

Thanks.

alexheretic:
Ideally the config shouldn't be provided by the client (vscode) at all, but allow individual users to provide it if they have a reason to override the behaviour of RLS.

I don't use vscode so I'm not sure if doing that is awkward for some reason.

Issue #340, opened 2018-06-14
Stop the RLS


Zaerei:
It would be very nice to be able to stop the RLS. Sometimes I need to use VS Code to resolve merge conflicts, or do other clerical work, or I intend to run the project from command line for a little bit. I'd prefer to not have to close VS Code or use another editor in order to keep recompiles from happening. On my home computer it's not a big deal, but on my laptop the RLS can eat a lot of resources (to the point things like "typing in the terminal" lags) and a command to just kill it would help a lot. 

This is especially problematic when dealing with git rebases and things like that because it tends to cause huge tree rebuilds of the project (especially in large workspaces) which constantly drop and restart because by the time it's done you've long moved on to the next step in a rebase (or whatever).

Apologies if there is a way to do this, but if VS code has a method for killing things beyond "terminate task" then I don't know where it is. Either way, I'd appreciate a more discover-able command along the lines of the "Restart the RLS" one.

repi:
I've been wanting to have something similar also when working on a laptop and needing to conserve battery, then it is not ideal to have RLS running constantly in the background and building and checking the code. So for that use case I would like to temporarily stop it and build/verify manually. Would be nice

Xanewok:
I'm somewhat torn on this - it's weird requesting the extension to stop working short of disabling the extension itself. I guess we could implement this; the "Restart the RLS" command should spawn it again.

mzabaluev:
Disabling the extension would also remove build target extraction and other useful features.
Stopping just the RLS helps to conserve power use and memory: a few times I had my laptop lock up in swapping/overheated state with several Code windows opened on big Rust projects, all churning RLS instances.

arjunepr:
Please get this fixed. I'm currently just learning Rust and I was having trouble with deleting the target directory from the command line... mysteriously, the target directory failed to be deleted every single time, and I realized what was going on after deleting it through vscode. The directory was regenerated, automatically, by the extension. So far I haven't found any setting to disable this behaviour. Hopefully, an option for the same would soon be added

MaulingMonkey:
Just ran into this poking through `.cargo\registry`... crev doesn't like the resulting 'unclean' repositories.  As a workaround, `--disable-extension` can be useful, e.g.:
```
cargo crev open sourcefile --cmd "\"%LOCALAPPDATA%\Programs\Microsoft VS Code\Code.exe\" --wait --disable-extension rust-lang.rust" --cmd-save
```

[...]

PR #682, opened 2019-10-17
Add start and stop commands for rls, add auto start rls config value

Yarn:
Addresses #340 
Adds "Stop the RLS" and "Start the RLS" commands.
Adds `rust.autoStartRls` setting.

`rust.autoStartRls` defaults to `true` keeping current behavior. If set to `false` opening files or adding projects to the workspace will not result in the RLS starting.

The existing auto-start behavior only triggered once per project until the extension is reloaded. As a result stopping the RLS will result in it staying stopped for a project until vscode is restarted/reloaded.

The start command can be used to start the RLS after it is stopped or if auto-start is disabled.


Yarn:
I'm still getting errors for `Error: command 'rls.update' already exists` in some cases (with `"rust-client.enableMultiProjectSetup": true,`). This stops RLS from being started.

With this change you can trivially trigger the bug by stopping RLS for a project where it is started then starting RLS in any other project (manually or not).
On master it required removing a project from the workspace then performing an action that would result in RLS being started.

The issue is that in `ClientWorkspace.stop` `commandsRegistered` is set to `false` but `this.disposables.forEach(d => d.dispose());` doesn't seem to be sufficient to allow `commands.registerCommand('rls.restart'...` to be called again without an error.

We can work around that by not setting `commandsRegistered` to `false` in `stop` if `rust-client.enableMultiProjectSetup` is set to `true`. I'll wait for confirmation if that's desired or should be part of a separate PR/have a less hacky solution.

rivy:
* fixes #340

Xanewok:
That's a great option and I'd love if we had that =) Do you think you could rebase this PR on top of the current master? Since this PR we've made some changes to command registration and multi-root handling, so hopefully it'll be a lot easier to apply now.

Commands are now registered globally but their behaviour alters depending on the `activeWorkspace`:
https://github.com/rust-lang/rls-vscode/blob/30b399dae0dde617af5c67b5cbe858079311ae07/src/extension.ts#L439-L454

Yarn:
Merge complete. I changed the setting to `rust-client.autoStartRls` since `rust.*` settings apparently get passed (sort of) directly to RLS.

PR #685, opened 2019-10-20
Use dynamic wait_to_build inferred by the RLS by default

Xanewok:
That was an oversight when we implemented dynamic `wait_to_build` on the RLS side - we forgot to update the default on the VSCode side. This substantially reduces the default latency on small-medium projects.

cc @alexheretic @janriemer
Closes #670

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ### Unreleased
 
+* Use dynamic `wait_to_build` in RLS by default rather than setting it to 1500ms
+
 ### 0.7.0 - 2019-10-15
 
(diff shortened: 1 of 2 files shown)


janriemer:
Thank you very much @Xanewok for taking charge of this issue and letting me know. I really appreciate it! :heart: 
This will now be a much better experience for rust-devs.

Xanewok:
I think it's rather uncontroversial and the dynamic wait_to_build has been implemented long enough, so I'll just go ahead and merge this.

Thanks for opening the initial PR and bringing this to my attention!

PR #696, opened 2019-11-02
Make spinner colorful and smoother

Gowee:
Before: ◐◓◑◒
Now: 🌕🌖🌗🌘🌑🌒🌓🌔

EpRSEAKiR6

The new spinner is smoother and colorful!

~~As an alternative, here is a more delightful spinner:~~

sxrtaSky6u

~~`const spinner = "😕 😟 🙁 😮 😯 😲 😳 🥺 😦 😧 😨 😰 😥 😢 😭 😱 😖 😣 😞".split(" "); //☹`~~

XD

### DIFF
--- a/src/spinner.ts
+++ b/src/spinner.ts
@@ -21,4 +21,4 @@ export function stopSpinner(message: string) {
 }
 
 let spinnerTimer: NodeJS.Timer | null = null;
-const spinner = ['◐', '◓', '◑', '◒'];
+const spinner = ['🌕', '🌖', '🌗', '🌘', '🌑', '🌒', '🌓', '🌔'];


Xanewok:
Thanks, that indeed looks better :)

BO41:
Nice idea, but it does not seem to work everywhere.

moon

Gowee:
@BO41 I suppose that it can be fixed by installing & configuring an emoji font. But it also suggests that the new spinner might not work out of the box. So I am afraid it is better to revert to the old one. What regret 🙁.

p-avital:
Or we could go overkill and have a setting for that 😃

BO41:
Shouldn't the status bar use the same font for all vscode installations?
Is there a way to check if the moons are supported on the installed font?
A separate setting is really kinda overkill

Xanewok:
Fixed in #771, will be using octicon font included in the vscode installation. Sorry about regression!

PR #701, opened 2019-11-16
Update prefix on spawn snippet

### LINKED ISSUE #669: Remove annoying unconditional thread::spawn snippet
The `thread::spawn` snippet seems to be suggested regardless of any context.

image

There is a very *very* narrow chance I actually want to use it. it usually just pollutes my autocompletion list.

imho it should be only suggested when the user types e.g. `thsp`.

Angarsk8:
Closes #669 

As already described in #669, the `thread::spawn` snippet gets triggered every time the `:` symbol is typed into the editor, which causes an unnecessary pollution in the autocomplete suggestions list, and can be quite annoying when dealing with things like: enums (to access its variants), modules (to access its definitions), structs (to call associated functions on it), "imports" on nested paths, etc. In general, anything that requires writing `::`, since as soon as the first `:` is typed, the suggestion for the `thread::spawn` snippet is retrieved.

In order to fix the aforementioned issue, I've updated the prefix `thread::spawn` to `thread_spawn` and `spawn`.

**Note**: Thanks for the great work you've done on this tool, I've been REALLY loving and enjoying learning rust using VSCode using this awesome integration of the RLS. Thanks again 🏅

### DIFF
--- a/snippets/rust.json
+++ b/snippets/rust.json
@@ -78,7 +78,10 @@
         "description": "Unwrap Option with if let"
     },
     "spawn": {
-        "prefix": "thread::spawn",
+        "prefix": [
+            "thread_spawn",
+            "spawn"
+        ],
         "body": [
             "thread::spawn(move || {",
             "\t$0",


fiahil:
Ping! Could someone click on the merge button? :)

danilaml:
Thanks. That was quite annoying to have when writing format specifiers in `println!` and the like. I hope this can get into the release soon.

PR #711, opened 2019-12-08
Use rust.sysroot option

### LINKED ISSUE #473: The `rust.sysroot` config option is ignored
`ClientWorkspace.getSysroot` asks `rustc` or `rustup` but doesn't check the config value.

This was confusing as I was trying to configure VSCode to deal with my weird environment.

majecty:
Rls has a `sysroot` option. rls-vscode also has the `sysroot` option. It seems that when we set the `sysroot` option in the vs-code rls, then the value should be delivered to the rls, but it was not.

Before this PR, rls-vscode wasn't using the sysroot option; instead, it was calculating the sysroot value by running `rustc --print sysroot` command.

In this PR, I changed the rls-vscode to use the sysroot value in the option. If a user doesn't set the sysroot option, vscode rls calculate the sysroot value using the `rustc --print sysroot` command.

This PR fixes #473.

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -79,6 +79,10 @@ export class RLSConfiguration {
     return this.configuration.get('rust-client.rustupPath', 'rustup');
   }
 
+  public get sysroot(): string {
+    return this.configuration.get('rust.sysroot', '');
+  }
+
   public get useWSL(): boolean {
     return this.configuration.get<boolean>('rust-client.useWSL', false);
   }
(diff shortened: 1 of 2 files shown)


majecty:
The Travis test failed in the Windows environment because the test code could not find "getmac.exe". The error seems not related to the changes I make. I manually run the `npm run test` command and it works in my WIndows machine.

majecty:
It's old. I have no time to follow it. I'll close this PR. Sorry.

PR #712, opened 2019-12-09
Specify Default Toolchain as Active Channel

Steven0351:
This patch enables the user to opt for the default toolchain in addition to stable, beta, or nightly. The default toolchain is currently only used as a fallback if the channel value is not present. The same effect could be achieved by deleting the settings.json entry, however I actually didn't figure that out until after perusing the source seeing how it's implemented. I think it would be helpful to be able to explicitly specify to use the default toolchain.

**Context**
I'm a new Rustacean and ran into the first occasion where I wanted to use a nightly feature. I had previously explicitly set the stable channel  when I first set the extension up in VS Code. When RLS complained about my nightly feature flag in my source, I popped over to the settings UI and switched the channel to nightly where it then complained that RLS wasn't available after trying to install it from the unstable channel. That led me into finding out that RLS isn't available in all versions of nightly Rust, so I installed an older nightly toolchain where RLS was available. I had expected this to work, but it still complained, and that's when I figured out the extension is using the nightly channel and not the nightly toolchain that was set as default. Even reading the text "By default, uses the same channel as your currently open project" it didn't really click to edit the settings.json file. I use VS Code as my daily driver for most of my coding and I constantly forget that the settings are backed by a json file when I'm twiddling with settings and it's not actually apparent through the UI that the stable channel isn't the default since it's the one showing up in the drop box by default.

### DIFF
--- a/package.json
+++ b/package.json
@@ -196,6 +196,7 @@
                         "null"
                     ],
                     "enum": [
+                        "default",
                         "stable",
                         "beta",
                         "nightly"
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -62,7 +62,7 @@ export class RLSConfiguration {
     configuration: WorkspaceConfiguration,
   ): string {
     const channel = configuration.get<string>('rust-client.channel');
-    if(channel === 'default' || !channel) {
+    if (channel === 'default' || !channel) {
       try {
         return getActiveChannel(wsPath, rustupConfiguration);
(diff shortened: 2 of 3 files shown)


Xanewok:
That makes sense and indeed, it's a shame that the VSCode settings aren't a bit more intuitive, especially for `null`able and enumerated properties :(

Thank you for your contribution and sorry it took so long!

PR #720, opened 2019-12-22
Add angular brackets to `autoClosingPairs`

### LINKED ISSUE #719: Add `<` and `>` to language-configuration.json
It would be quite convenient (especially with generics), if vscode would automatically insert `>` if `<` is typed (similar to what is done with `{` and `}`)

https://github.com/rust-lang/rls-vscode/blob/15f14cbcbccf8ea21dee788a013fa9a205bcb231/language-configuration.json#L12

I think adding a new item to the list would be enough? `{ "open": "<", "close": ">" }`

The plugin for atom does this too
https://github.com/rust-lang/atom-ide-rust

Luro02:
I am currently switching from github's [atom](https://atom.io/) editor to vscode and one thing, that I immediately noticed, was that, when you type `<` there will no `>` added automatically, like it is done with `{` and `}`. At first, I suspected, that this was a global setting, but after a bit more testing and a post on [`r/vscode`](https://www.reddit.com/r/vscode/comments/ednx1l/is_it_possible_to_automatically_add_a_closing/) I found out, that this seems to be an issue with the language extension.

### Open questions
Should this be a separate setting for the language extension? If so how do I make this a setting? (I never did anything with vscode extensions)

fixes #719

### DIFF
--- a/language-configuration.json
+++ b/language-configuration.json
@@ -13,6 +13,7 @@
        { "open": "{", "close": "}" },
        { "open": "[", "close": "]" },
        { "open": "(", "close": ")" },
+       { "open": "<", "close": ">" },
        { "open": "\"", "close": "\"", "notIn": ["string"] },
        { "open": "/**", "close": " */", "notIn": ["string"] },
        { "open": "/*!", "close": " */", "notIn": ["string"] }


radu-matei:
This works great for generics, but whenever you're trying to compare two things (`x < y`), you get the auto-closing pair.r

Luro02:
I am aware of this problem, the same thing happens with the atom version, but I think, that you write far more `<>`, than right hand comparisons. Therefore I was asking to make this a separate setting.

Xanewok:
I don't have any metrics but from my experience there's a lot more generic code being written than the comparison (weirdly?):
- to introduce generics, e.g. `impl<T>`
- to use the types in type definition, e.g. `enum Abc<T>`, `struct Abc<T>`, `Trait<Assoc = SomeType>`
- when using fully qualified paths, e.g. `<Self as some::Trait>::method()`
- when using explicit types in generic functions e.g. `something.parse::<u32>()`

so I'd be inclined to appeal to the (hopefully) more common case :) I hope it's not going to be a lot of inconvenience - I'm open to changing that if people find it more harmful than helpful.

Thanks for your contribution!

towry:
@Xanewok https://github.com/rust-lang/vscode-rust/issues/798#issuecomment-767442725

?????????

PR #727, opened 2020-01-10
Update README.md in order to point correctly to Preferences

andresgallego:


Xanewok:
This doesn't seem to be correct; under Linux I still access it via File > Preferences > Settings. What system/VSCode version are you using?

Xanewok:
I'll close this due to inactivity.

Issue #419, opened 2018-09-04
Handle cases when the outer workspace is not a rust project


samrg472:
I'm using Neon to create a NodeJS native module with Rust. The project layout consists of an npm project as the root and a subfolder "native" that has the rust project.

In VSCode I create a workspace with 2 folders:
- neon-js/
- neon-js/native

The function `getOuterMostWorkspaceFolder` presents an issue because the outer most folder is not a rust project.
https://github.com/rust-lang-nursery/rls-vscode/blob/096974566afe1ab34ae0f868c7dd6a127f41b059/src/extension.ts#L85-L97

I resolved this problem locally by checking if the outer most folder is a rust project. I'm not sure if this will cause problems with something else or if there's a better way, but this solution is working for me.

```diff
diff --git a/src/extension.ts b/src/extension.ts
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -90,7 +90,13 @@ function getOuterMostWorkspaceFolder(folder: WorkspaceFolder): WorkspaceFolder {
             uri = uri + '/';
         }
         if (uri.startsWith(element)) {
-            return workspace.getWorkspaceFolder(Uri.parse(element)) || folder;
+            const ws = workspace.getWorkspaceFolder(Uri.parse(element)) || folder;
+            for (const f of fs.readdirSync(ws.uri.fsPath)) {
+                if (f === 'Cargo.toml') {
+                    return ws;
+                }
+            }
+            return folder;
         }
     }
     return folder;
```

Aidiakapi:
A similar usage scenario for this is with WebAssembly. The template project for this scenario (https://github.com/rustwasm/rust-webpack-template) also sets it up so a workspace root will be the webpack project, and the Rust project is stored under `./crate`.

I will give the proposed diff a try.

petkahl:
It seems like this is a duplicate of #222 which has a PR filed recently, but hasn't been merged.

joepie91:
The mentioned PR in #222 has been merged, but I'm still running into this issue with a Neon project; the error message is slightly different for me than the one in #222, though:

![selection_936](https://user-images.githubusercontent.com/1663259/53337172-fbdf2800-3900-11e9-9ae9-bcf13870b9b2.png)

This occurs in a blank new Neon project, created with neon-cli 0.20. The folder structure is as follows:

[...]

PR #759, opened 2020-04-05
Fix workspace path when multiProjectSetup is enabled

### LINKED ISSUE #713: RLS fails to start in multi-project setup on non-OS drive
The VS Code error message: ```Could not spawn RLS: `spawn C:\WINDOWS\system32\cmd.exe ENOENT` ```

Setting `"rust-client.enableMultiProjectSetup"` to `false` in settings.json prevents the issue.

Only occurs when running in a workspace folder on a different disk drive than the Windows OS drive.
In my case, Windows is installed to C:\, and I'm attempting to work in D:\projects\foo.

Example directory structure:
- D:\projects\foo
  - bar\
    - src\
      - main.rs
    - Cargo.toml
  - baz\
    - src\
      - main.rs
    - Cargo.toml

https://github.com/rust-lang/rls-vscode/blob/c85bf69f3a00184ee4600da34e6387178eba754c/src/extension.ts#L454
I noticed `cwd` here is set to `"\projects\foo"`, without the drive letter.

`settings.json`:
```json
"rust-client.enableMultiProjectSetup": true
"rust-client.channel": "stable"
```
```
PS D:\projects\foo> rustup -V
rustup 1.20.2 (13979c968 2019-10-16)

PS D:\projects\foo> rustup show
Default host: x86_64-pc-windows-msvc
rustup home:  C:\Users\...\.rustup

installed toolchains
--------------------

stable-x86_64-pc-windows-msvc
nightly-x86_64-pc-windows-msvc

active toolchain
----------------

stable-x86_64-pc-windows-msvc (default)
rustc 1.39.0 (4560ea788 2019-11-04)

PS D:\projects\foo> code -v
1.40.2
f359dd69833dd8800b54d458f6d37ab7c78df520
x64

PS D:\projects\foo> code --list-extensions --show-versions | rg rust
rust-lang.rust@0.7.0
```

jannickj:
When `enableMultiProjectSetup=true` the following bugs will be fixed

* Fix a minor bug where workspace folder was stripped from short-circuit check
* Use correct pathing format (Fixes pathing for windows)
* Fix document selector for windows. (windows wouldn't trigger rls hover/actions)
* Fix `cargo build` as it would previously point to the root folder (pointed out by @jli in #419).

The branch has been tested on *windows 10* & *ubuntu 18.04*

Will fix #713

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -310,6 +310,7 @@ class ClientWorkspace {
       const ws =
         multiProjectEnabled && activeWorkspace ? activeWorkspace : this;
       await ws.stop();
+      commandsRegistered = true;
       return ws.start(context);
     });
     this.disposables.push(restartServer);
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -208,7 +208,7 @@ class ClientWorkspace {
     };
 
     const pattern = this.config.multiProjectEnabled
-      ? `${this.folder.uri.path}/**`
+      ? `**${this.folder.uri.path}/**`
       : undefined;
     const collectionName = this.config.multiProjectEnabled
       ? `rust ${this.folder.uri.toString()}`
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -207,8 +207,14 @@ class ClientWorkspace {
       return this.makeRlsProcess();
     };
 
+    // Something else is put front of files when pattern matching that prevents the windows version from picking up the files
+    // This should be safe as the uri is a absolute path that includes the drive + colon
+    // i.e. a pattern would become "**/c:/some/path**" and since colon is reserved only the root can ever contain it.
+    const isWin = process.platform === 'win32';
+    const windowsHack = isWin ? '**' : '';
+
     const pattern = this.config.multiProjectEnabled
-      ? `**${this.folder.uri.path}/**`
+      ? `${windowsHack}${this.folder.uri.path}/**`
       : undefined;
 
     const collectionName = this.config.multiProjectEnabled
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -210,8 +210,9 @@ class ClientWorkspace {
     const pattern = this.config.multiProjectEnabled
       ? `**${this.folder.uri.path}/**`
       : undefined;
+
     const collectionName = this.config.multiProjectEnabled
-      ? `rust ${this.folder.uri.toString()}`
+      ? `rust ${this.folder.uri.path}`
       : 'rust';
     const clientOptions: LanguageClientOptions = {
       // Register the server for Rust files
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -212,7 +212,7 @@ class ClientWorkspace {
       : undefined;
 
     const collectionName = this.config.multiProjectEnabled
-      ? `rust ${this.folder.uri.path}`
+      ? `rust ${this.folder.uri.toString()}`
       : 'rust';
     const clientOptions: LanguageClientOptions = {
       // Register the server for Rust files
--- a/package.json
+++ b/package.json
@@ -224,7 +224,7 @@
                 "rust-client.enableMultiProjectSetup": {
                     "type": "boolean",
                     "default": false,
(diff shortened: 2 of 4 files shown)


jannickj:
There seems to be build server problem but I don't think it's related to my branch. @nrc / @Xanewok is that an incorrect assumption?

jannickj:
Applied the security fixes, just to get a building branch :), I can remove them if you'd rather do it yourself

Xanewok:
Sorry for taking so long. Would you mind rebasing against master again? It'd be great to also have some kind of a regression test for this :)

Xanewok:
> Applied the security fixes, just to get a building branch :), I can remove them if you'd rather do it yourself

Just did that in the meantime, so it'd be good to skip the relevant commits as well.

src/extension.ts
```diff
@@ -90,7 +90,7 @@ function whenOpeningTextDocument(
     return;
   }
 
-  const folderPath = folder.uri.toString();
+  const folderPath = folder.uri.path;
```
Xanewok, line 93: Does this change give us anything? I feel like we shouldn't use unix-only path if there's not a clear reason; If we should, then we ideally need to somehow document the rationale as well

jannickj: I think it was just to make hashing identical across systems. I can't remember if it was actually necessary (probably not, so I'll remove it)

jannickj: probably not, I think it was just to make hashing identical across systems.

src/extension.ts
```diff
@@ -208,10 +208,11 @@ class ClientWorkspace {
     };
 
     const pattern = this.config.multiProjectEnabled
-      ? `${this.folder.uri.path}/**`
+      ? `**${this.folder.uri.path}/**`
       : undefined;
+
     const collectionName = this.config.multiProjectEnabled
-      ? `rust ${this.folder.uri.toString()}`
+      ? `rust ${this.folder.uri.path}`
```
Xanewok, line 215: Is it the same instance as https://github.com/rust-lang/rls-vscode/pull/759#discussion_r409780233?

jannickj: yes.. :sweat_smile:

jannickj: The only place where it is necessary is for the pattern

Xanewok: Thanks for responding so quickly and making me understand the changes better!

src/extension.ts
```diff
@@ -208,8 +208,9 @@ class ClientWorkspace {
     };
 
     const pattern = this.config.multiProjectEnabled
-      ? `${this.folder.uri.path}/**`
+      ? `**${this.folder.uri.path}/**`
```
Xanewok, line 211: So I'm still wondering about this change... What exactly is the output for both `.path` and `.fsPath` and why do we need to prefix that with a wildcard?

Is the wildcard supposed to make the pattern drive-letter-agnostic, e.g. it'd match both `C:\Some\Path` and `D:\Some\Path` with a `**/some/path/**` pattern?

jannickj: right that's a really good point.

for fsPath it's what you would expect just standard operative system paths
windows: `c:\Some\Path`
linux: `/some/path`

path is what is interesting (it's basically the path part of a FILE uri):
windows: `/c:/Some/Path` with the file uri: `file:///c:/Some/Path`
linux: `/Some/Path` with the file uri: `file:///Some/Path`

Now the pattern matcher uses file uri paths for matching, my guess is they wanted to have a OS agnostic pattern matcher.

The `**` is because windows puts something else infront of the files when matching and I have no clue what however the only way to get around it was `**` which will catch it.

jannickj: Maybe a better solution would be to only apply this for linux

Xanewok: Thanks for clarifying! What do you mean by something else? Do you mean like `c%3A` (which is percent-encoded `c:`)?

jannickj: No, I meant they put something `[something]/c:/...` I don't know what that something is and I couldn't figure it out, I thought it was fine however but I understand wanting to fix it the correct way :).

Xanewok:
Thank you! I don't want to block this PR further, really appreciate your time and effort to land these fixes 🙇

PR #763, opened 2020-04-18
Fix support for multiple VSCode workspaces

Xanewok:
Closes #601.

This is based on #601 (thanks @zkat!) and extended a bit to fix our integration with custom signature helper.

The gist is of it is to globally register the commands so there are no more exceptions once we try to spin up another instance, which prevented us from doing so, earlier.

The fix for signature helper means we have to drop some checks for `multiProjectEnabled` and unconditionally use the logic as if the config was turned on, which leaves me wondering if we should just enable it unconditionally...

Tested and it seems to be working both for multiple VSCode workspaces and it didn't regress the multiple projects per a single workspace.

@zkat @jannickj do you mind taking a look at this?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ### Unreleased
 
+* Fix support for multiple VSCode workspaces
+
 ### 0.7.2 - 2020-04-17
 
 * Fix a bug where rustup didn't install all of the required components for the RLS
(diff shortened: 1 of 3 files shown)


jannickj:
One thing I would check is if the tasks: cargo build works (I had a bunch of issues with that)

Otherwise looks good to me, and I agree that we could probably set `multiProjectEnabled` as on by default (and eventually remove it), the reason why it was behind a feature flag in the first place was it was treading very untested areas and caused a lot of problems.

Xanewok:
Double-checked on both Windows and a Ubutu box and it seems to be working as expected, so I'll merge this.  Thanks for taking a look!

PR #774, opened 2020-04-26
Fix document selectors in remote scenarios

### LINKED ISSUE #706: Does nothing on Windows 10
I tried to use rls-vscode with VSCode, but failed horribly. 

I actually re-installed the whole toolchain to make it work, but the extension does...well...nothing. It compiles my example file (I'm seeing the RLS stuff in the footer), but other than that...nothing.
The code is littered with intentional errors, but nothing gets highlighted. Code completion is not working as well.

I did not change any config and installed RLS from the popup that appears during first run.

I'd be happy to provide any additional information.

I'm using Windows 10 64-Bit, VSCode 1.40.1, rustc/cargo 1.39.0, rustup was installed today.

Thanks for all the work you guys do, I hope you can solve my issue, too.

### LINKED ISSUE #773: Couldn't use the extension if I'm in a remote development environment.
Hi,
I'm trying to develop rust program on my server just like I use C language on my server with the remote development extension. But I find out the rls extension doesn't work on my case. I'm wondering if rls is not supported in remote development, or I just missed something.

Thanks!

Xanewok:
Closes #773
Closes #706 (apart from autocompletion, this should be improved on separately)

cc @jannickj since this touches on patterns again - turns out `RelativePattern`s don't work properly in remote sessions, so I fell back on using fs paths from URIs.

Tested on both Xubuntu 19.10 and Windows 10 (Remote WSL + SSH on Xubuntu), including other folders like D:\ and so far so good.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -8,7 +8,6 @@ import {
   ExtensionContext,
   IndentAction,
   languages,
-  RelativePattern,
   TextEditor,
   Uri,
   window,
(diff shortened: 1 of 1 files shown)


Xanewok:
I'll merge this to cut a release now before Monday hits in full swing.

jannickj:
Patterns is a real headache, I hope you've found the secret command :D

astone2014:
I tested RLS 0.7.4 and it works much better in WSL2.
Autocomplete still doesn't work for me except for some things. Like a function I write.
No rust language auto complete for basic types.

Sadly I think I'm going to have to stick to Windows for now.

jannickj:
@astone2014 Yeah, I think it's probably to do with how RLS is running, it would probably make sense to implement a proper WSL setup.

PR #784, opened 2020-04-29
Try running the tests without explicit xvfb wrapper

Xanewok:


Xanewok:
This removes the need to use a custom action but it seems that `vscode-test` doesn't properly pass the exit code to the main invocation - seemingly the tests are flagged as passing but there seems to have been a regression introduced in #777

PR #796, opened 2020-05-19
Add String::from() snippet

Filius-Patris:
This snippet listens on `strfrom` and writes `String::from("")` (with the cursor between the two quotation marks).

I think this is a really helpful snippet, because I find myself often typing out `String::from()`.


gmr-em:
The rest of the file uses spaces so maybe replace your tabs to adhere to that convention :)

PR #802, opened 2020-05-28
Add tasks.json Support

jcdickinson:
`resolveTask` is now used by VSCode to resolve a task.json task to an execution. This commit adds support for this functionality, by porting changes made to the RA plugin: https://github.com/rust-analyzer/rust-analyzer/pull/4096

The `cargo` type should now work in the same way to other languages available in VSCode. E.g.

```json
{
	"type": "cargo",
	"command": "build",
	"args": [
		"--target",
		"x86_64-pc-windows-gnu"
	],
	"problemMatcher": [
		"$rustc"
	],
	"group": {
		"kind": "build",
		"isDefault": true
	},
	"label": "rust: cargo build1"
}
```

### DIFF
--- a/fixtures/.vscode/tasks.json
+++ b/fixtures/.vscode/tasks.json
@@ -1,0 +1,21 @@
+{
+	"version": "2.0.0",
+	"tasks": [
+		{
+			"type": "cargo",
+			"command": "build",
+			"args": [
+				"--target",
+				"x86_64-pc-windows-gnu"
+			],
+			"problemMatcher": [
+				"$rustc"
+			],
+			"group": {
+				"kind": "build",
+				"isDefault": true
+			},
+			"label": "custom build task"
+		}
+	]
+}
--- a/package.json
+++ b/package.json
@@ -115,14 +115,31 @@
         "taskDefinitions": [
             {
                 "type": "cargo",
-                "properties": {
-                    "subcommand": {
-                        "type": "string"
-                    }
-                },
                 "required": [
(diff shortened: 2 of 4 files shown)


Xanewok:
I'm not sure I follow - the existing implementation already supports custom Cargo tasks in tasks.json without `resolveTasks`. Do you mean supporting additionally the `args` part of the task definition?

Xanewok:
And sorry for looking at this so late - let me know if you're still interested in landing this.

jcdickinson:
I'm still happy to help get this landed.

> I'm not sure I follow - the existing implementation already supports custom Cargo tasks in tasks.json without `resolveTasks`. Do you mean supporting additionally the `args` part of the task definition?

There is more info in this RA issue https://github.com/rust-analyzer/rust-analyzer/issues/3983. I'm not sure exactly what triggers the symptom (possibly using tasks with names other than the built in tasks). I did test both plugins at the time, and both had this problem, both being fixed by implementing `resolveTask`.

Either way, `args` is also pretty useful and [I was using it at the time.](https://github.com/jcdickinson/racemus/blob/master/.vscode/tasks.json#L32)

huntc:
> Either way, `args` is also pretty useful and [I was using it at the time.](https://github.com/jcdickinson/racemus/blob/master/.vscode/tasks.json#L32)

Would just like to chime in and add my vote for being able to pass args. Spent a good hour looking for this option today. Thanks.

PR #816, opened 2020-07-13
Move rust-analyzer extension into this repository

matklad:
r? @Xanewok 

This just copy-pastes (with history) rust-analyzer plugin into `./rust-analyzer` directory, so that it's easier to move functionality into the merged plugin. That is, I expect us to drain this `./rust-analyzer` directory after this RP


lnicola:
I guess `lodash` was finally fixed? CI fails because of `npm audit`.

Xanewok:
@matklad awesome, thanks! let's merge this now and sort out lodash dep separately

PR #817, opened 2020-07-13
Bump lodash

Xanewok:
To fix the low severity `npm audit` issue.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -766,9 +766,9 @@
             }
         },
         "lodash": {
-            "version": "4.17.15",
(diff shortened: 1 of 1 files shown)


Xanewok:
@matklad it seems the Windows CI is red due to prettier changes? Wild guess is because it's due to EOL? Would be great if we could figure out if we can cheaply make it pass again for the time of gradual merging but I think we can also work around that

matklad:
We don't use prettier, we use tsfmt. So I assume that prettier legitimately complains about different formatting. So I would guess that the problem is that we actually only run prettier on Windows, due to different shell glob expansion?

PR #826, opened 2020-07-22
Fix snippetTextEdits applying to other files

TimoFreiberg:
vscode.window.visibleTextEditors only contains editors whose contents
are being displayed at the moment, so the previous logic only worked if
the other file for which a snippetTextEdit is being received was visible
in a separate split.

Ported from rust-analyzer: https://github.com/rust-analyzer/rust-analyzer/pull/5480


Xanewok:
Thanks!

PR #827, opened 2020-07-28
Fix: respect `rust-client.disableRustup` setting when using rust-anyzer

gentoo90:
Currently `rust-analyzer` fails to start if `rustup` is missing even if it's disabled in the settings


Xanewok:
Thank you so much and sorry for the delay!

PR #833, opened 2020-08-09
rust-analyzer: Listen for new "roots scanned" message to indicate progress

tchebb:
As of rust-analyzer/rust-analyzer@dad1333b48c3, the `rustAnalyzer/startup` message has been replaced with a functionally-equivalent[1] `rustAnalyzer/roots scanned` message. Since we're still looking for the old message, which is no longer sent at all, our progress UI hangs on "Starting" when using rust-analyzer. Listen for the new message instead to fix this.

[1] At least, it appears functionally equivalent to me, but I don't know anything about rust-analyzer's internals so I could be wrong.

### DIFF
--- a/src/rustAnalyzer.ts
+++ b/src/rustAnalyzer.ts
@@ -285,7 +285,7 @@ async function setupGlobalProgress(client: lc.LanguageClient) {
     if (newState === lc.State.Starting) {
       await client.onReady();
 
-      const RUST_ANALYZER_PROGRESS = 'rustAnalyzer/startup';
+      const RUST_ANALYZER_PROGRESS = 'rustAnalyzer/roots scanned';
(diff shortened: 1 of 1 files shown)


Xanewok:
@matklad does this look good for you?

matklad:
We indeed changed the name of this thing, so in this sense the fix seems reasonable. But, in terms of overall idea of progress handing, rust-analyzer does proper progress reporting itself since some time ago, so this whole thing should not really be necessary. 

Rather, the status messages are now used to display progress-bar item:

image

matklad:
I think a good move is to merge, and refactor away separately later .

Xanewok:
Okay, makes sense, thanks for more context! Let's merge this, then.

PR #956, opened 2022-05-21
Deprecate `rust-lang.rust`

Veykril:
Adds a notice to the extension name and readme as well as triggering a notification pop on start linking to the rust-analyzer marketplace pages.

image

cc https://github.com/rust-lang/rls/pull/1776

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,9 @@
-# Rust support for Visual Studio Code
+# Rust support for Visual Studio Code (deprecated)
 
 [![](https://vsmarketplacebadge.apphb.com/version/rust-lang.rust.svg)](https://marketplace.visualstudio.com/items?itemName=rust-lang.rust)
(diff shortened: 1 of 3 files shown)


bjorn3:
Could a similar message be put into rls itself for non-vscode users? That message should probably get tweaked a bit to not refer to the vscode extension id's though.

Veykril:
Ye given rls is lsp we should be able to send a notification via the LSP layer as well.

lnicola:
We might want a "don't show again" option, since some people will surely want to keep using RLS.

Xanewok:
This will be updated shortly via https://github.com/rust-lang/vscode-rust/issues/927#issuecomment-1140302139 so I'd like to wait a couple of days for the official support, instead.

ehuss:
@Xanewok Not sure if you've been following, but we have been discussing the deprecation at https://rust-lang.zulipchat.com/#narrow/stream/301329-t-devtools/topic/rls.20deprecation.

I tested the deprecation system added in the insiders build, and AFAICT it does not provide any sort of notification to the user. All I can see is that it grays out the extension in the search page, and removes the Install button.  

And, FWIW, this PR shouldn't move forward until some time after the public notification.  I am planning on posting a PR for that today.

lnicola:
`vscode.rust` is already marked as deprecated on the Marketplace, with a pretty decent UX:

image

And in the meanwhile, https://blog.rust-lang.org/2022/07/01/RLS-deprecation.html came out.

